PR #3, opened 2026-05-11
fix: extend Metal watchdog eval guards to all Apple Silicon Macs

colinbdesign:
*Note: I'm not a developer — this fix was found by Claude Code while diagnosing a crash on my machine. I'm sharing it back in case it helps others. The fix may not be how you'd approach it, but it resolved the crash for me.*

## Problem

On M2 Max 64 GB (macOS 15.7.3 Sequoia), every generation attempt crashes within ~10 seconds:

```
libc++abi: terminating due to uncaught exception of type std::runtime_error:
[METAL] Command buffer execution failed: Internal Error (0000000e:Internal Error)
```

The same crash is reported on another M2 Max 64 GB machine running Sonoma (phosphene issue #3).

## Root cause

The existing Metal watchdog guards in `base_encoder.py` and `embeddings_connector.py` (added in commit 3431205) gate per-layer/per-block `mx.eval()` calls behind a `memory_size <= 48 GB` threshold. On a 64 GB machine this condition is false, so Gemma's 48-layer forward pass and the connector blocks accumulate as one giant lazy graph and submit to Metal as a single command buffer — exceeding the ~10 s watchdog deadline.

The threshold was validated at 384×256×9 frames, where the lazy graph completes within the window. At production resolutions (640×480+, 121 frames) it does not.

## Fix

- `base_encoder.py`: removed the memory-size gate; `LTX2_GEMMA_EVAL_EVERY` now defaults to `1` (per-layer) on all devices. Set `LTX2_GEMMA_EVAL_EVERY=0` to restore full lazy-graph behaviour.
- `embeddings_connector.py`: same — per-block eval is now the default for all devices, honouring `LTX2_GEMMA_EVAL_EVERY=0` to opt out.
- `model.py`: added `LTX2_DIT_EVAL_EVERY` (default 8) — flushes the DiT block loop every 8 layers on the non-streaming path, for the same reason.

## Tested on

M2 Max 64 GB, macOS 15.7.3, Q4 distilled model, Quick preset (640×480, 121 frames): generation completes successfully in ~199 seconds.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -851,20 +851,21 @@ End-to-end run on M2 Pro 32 GB, dev model + HDR LoRA fused, q8:
 
 ---
 
-## Metal Watchdog Mitigation (auto, memory-gated)
+## Metal Watchdog Mitigation
 
 The macOS GPU watchdog (`kIOGPUCommandBufferCallbackErrorImpactingInteractivity`, ~10 s per Metal command buffer) trips when:
 
 1. A single command buffer takes too long (one giant lazy-graph dispatch).
 2. Many small command buffers queue behind system processes (mds_stores, knowledgeconstructiond, Spotlight, Siri).
 
-LTX-2 mitigates both automatically on `<=48 GB` Macs by inserting `mx.eval` at strategic points so each command buffer is small enough to fit one watchdog window but few enough that queue contention doesn't dominate:
+LTX-2 mitigates both by inserting `mx.eval` at strategic points so each command buffer is small enough to fit one watchdog window but few enough that queue contention doesn't dominate. These guards apply on **all Apple Silicon Macs** — the previous 48 GB threshold was empirically wrong; M2 Max 64 GB machines exhibit `MTLCommandBufferErrorInternal` (code 14) crashes without them:
 
-- **Gemma forward**: per-layer eval (48 layers × ~100 ms each). Override via `LTX2_GEMMA_EVAL_EVERY=N` (default `1` on ≤48 GB, `0` on bigger Macs).
-- **TextEmbeddingProjection**: per-output-projection eval (splits the 188160→4096 video matmul from the 188160→2048 audio matmul). No-op on >48 GB.
-- **Embeddings1DConnector**: per-block eval (8 video + 8 audio blocks). No-op on >48 GB.
+- **Gemma forward**: per-layer eval (48 layers × ~100 ms each). Override via `LTX2_GEMMA_EVAL_EVERY=N` (default `1`).
+- **TextEmbeddingProjection**: per-output-projection eval (splits the 188160→4096 video matmul from the 188160→2048 audio matmul).
+- **Embeddings1DConnector**: per-block eval (8 video + 8 audio blocks).
+- **LTX DiT block loop**: eval every `LTX2_DIT_EVAL_EVERY` blocks (default `8`). Splits the 48-block forward into 6 command buffers of ~6 blocks each (~1–2 s/buffer), well within the watchdog window.
 
-`>48 GB` Macs (Mac Studio, M-series Ultra) keep full lazy-graph pipelining for max throughput.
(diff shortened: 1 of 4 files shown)


dgrauet (commented):
Thanks for the careful write-up — solid diagnostic, reproducible on two independent M2 Max 64 GB machines, and the fix is empirically validated. Happy to take this. A few polish items before merge:

## Required

**1. Hoist the per-forward env reads out of `__call__`** (`transformer/model.py`)

`_dit_eval_every = int(_os.environ.get(...))` and `_mx_eval = getattr(mx, "eval")` currently re-execute on every forward — that's hundreds of calls per generation. Move them to module scope:

```python
# At the top of model.py
import os as _os
_DIT_EVAL_EVERY = int(_os.environ.get("LTX2_DIT_EVAL_EVERY", "8"))
_mx_eval = getattr(mx, "eval")  # noqa: B009
```

then reference `_DIT_EVAL_EVERY` / `_mx_eval` directly inside `__call__`. Same pattern as `base_encoder.py` already uses for `eval_every`.

**2. Document the new knob + perf trade-off in `CLAUDE.md`**

The "Metal Watchdog Mitigation" section currently says the per-layer / per-block evals are *"No-op on >48 GB Macs"* — that's no longer true after this PR. Please update that section to:
- Drop the 48 GB threshold language (it's empirically wrong).
- Add `LTX2_DIT_EVAL_EVERY` to the env-var list (default 8, set to 0 to disable).
- Note that Mac Studio M-series Ultra users who never see the watchdog crash may set `LTX2_GEMMA_EVAL_EVERY=0` and `LTX2_DIT_EVAL_EVERY=0` to recover full lazy-graph pipelining.

## Nice to have (non-blocking)

**3. Quick perf measurement on >48 GB**

The previous gate was deliberate — on machines that didn't crash, the per-block syncs cost a few % per step. Now they apply universally. If you have a minute on your M2 Max 64 GB, a quick comparison of `LTX2_DIT_EVAL_EVERY=8` (new default) vs `LTX2_DIT_EVAL_EVERY=0` (old behaviour, on machines that don't crash) at your validated config (640×480×121) would let us quote an overhead number in the changelog. Not required for merge.

---

Mechanically the patch is sound. The DiT guard in particular is a real gap we didn't cover — Gemma + connector were the only paths gated before, but the DiT block loop is by far the largest workload per forward, so it was the most likely watchdog offender all along.

Disclaimer noted on the "not a developer" front — no worries, the diagnostic is genuinely good. Once 1+2 land, I'll merge.

colinbdesign:
Both required items addressed in d560057:

**1. Module-scope hoist (`model.py`)** — `import os as _os`, `_DIT_EVAL_EVERY`, and `_mx_eval` moved to module scope, pattern matches `base_encoder.py`. The in-call copies are removed.

**2. CLAUDE.md updated** — 48 GB threshold language removed, `LTX2_DIT_EVAL_EVERY` added to the env-var list, note added for Mac Studio / M-series Ultra users to recover full lazy-graph pipelining via `LTX2_GEMMA_EVAL_EVERY=0` + `LTX2_DIT_EVAL_EVERY=0`.

---

**Benchmark — `LTX2_DIT_EVAL_EVERY=8` vs `=0` (M2 Max 64 GB, 640×480×121 frames, 8 steps, Quick preset, seed=42)**

| Config | Wall time |
|---|---|
| `LTX2_DIT_EVAL_EVERY=8` (new default) | 202.0 s |
| `LTX2_DIT_EVAL_EVERY=0` (full lazy graph) | 199.9 s |
| **Overhead** | **~2 s / ~1%** |

The 6 extra `mx.eval()` sync points (48 blocks ÷ 8) add ~2 seconds at this resolution — well within measurement noise and a reasonable trade for watchdog safety. On machines that don't crash, the cost of enabling guards is negligible.

dgrauet (approved):
All required items addressed cleanly, plus the bonus benchmark (~1% overhead on M2 Max 64 GB) — well worth the watchdog safety. Thanks!

PR #16, opened 2026-05-13
fix(distilled): I2V+LoRA crashes and Metal watchdog on 64 GB Apple Silicon Macs

colinbdesign:
*Note: I'm not a developer. These fixes were found and written by Claude Code while diagnosing crashes on my machine. I'm sharing them back in case they help. Claude Code is also writing this message on my behalf.*

---

## Background

PR #3 (submitted by me, merged by you) fixed the Metal watchdog crashes — thank you for reviewing and merging that. After that landed, I kept investigating further crashes specific to I2V and LoRA-enabled generation on my M2 Max 64 GB Mac. This PR covers three related issues all rooted in the 64 GB memory environment.

## Fix 1 — I2V tiling artefacts after removal of `ImageToVideoPipeline`

**Symptom:** Every I2V generation showed systematic colorful tiling/mosaic artefacts across all frames after frame 0.

**Root cause:** After upstream commit `493aec2` removed `ImageToVideoPipeline`, callers checking `hasattr(pipe, "generate_from_image")` fell back to `generate_two_stage()`, which runs stage 1 at half resolution. For 640×480 that is H=7, W=10 = 1120 tokens. The distilled model cannot produce coherent I2V conditioning at such small spatial dimensions when the reference image is encoded at full resolution — the token layout is incompatible.

**Fix:** Add `generate_from_image()` to `DistilledPipeline`. Single-stage at full target resolution (H=15, W=20 = 4800 tokens for 640×480), using `DISTILLED_SIGMAS` (8 steps, no CFG) — mirroring the removed method.

## Fix 2 — Metal watchdog crashes during I2V generation

**Symptom:** `MTLCommandBufferErrorInternal` code 14 on the first denoise command buffer.

**Root cause:** At 4800 tokens (full-res), `LTX2_DIT_EVAL_EVERY=8` produces command buffers that exceed the 10-second watchdog — O(N^1.5) scaling plus cold Metal shader compilation for dequantize/add/requantize kernels.

**Fix (three guards):**
- Pre-materialise DiT weights block-by-block before the denoise loop (compiles Metal kernels upfront, avoids >30 GB peak from fused lazy chains)
- Override `LTX2_DIT_EVAL_EVERY` to `min(2, default)` for the duration of `generate_from_image`, restored in a `finally` block
- Flush the initial noised state and VAE conditioning tensors into their own Metal command buffers before the denoise loop

The same noised-state flush is applied to both stages of `generate_two_stage()` where the same condition can trigger.

## Fix 3 — LoRA silently ignored for T2V

**Symptom:** T2V with a LoRA produced identical output to T2V without any LoRA.

**Root cause:** `DistilledPipeline.load()` loaded the DiT without checking `_pending_loras`. T2V goes through `generate_two_stage() → self.load()`, so any LoRA set by the caller before the first job was never fused.

**Fix:** After loading the DiT in `load()`, check `_pending_loras` and apply the same post-load block-by-block fusion used in `generate_from_image()`.

---

All three issues were reproducible on M2 Max 64 GB. I don't have access to a smaller-RAM Mac to test the non-64 GB paths, but the changes are guarded to only activate when `_pending_loras` is set or when `generate_from_image` is explicitly called, so the default two-stage T2V path is unaffected.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -851,20 +851,21 @@ End-to-end run on M2 Pro 32 GB, dev model + HDR LoRA fused, q8:
 
 ---
 
-## Metal Watchdog Mitigation (auto, memory-gated)
+## Metal Watchdog Mitigation
 
 The macOS GPU watchdog (`kIOGPUCommandBufferCallbackErrorImpactingInteractivity`, ~10 s per Metal command buffer) trips when:
 
 1. A single command buffer takes too long (one giant lazy-graph dispatch).
 2. Many small command buffers queue behind system processes (mds_stores, knowledgeconstructiond, Spotlight, Siri).
 
-LTX-2 mitigates both automatically on `<=48 GB` Macs by inserting `mx.eval` at strategic points so each command buffer is small enough to fit one watchdog window but few enough that queue contention doesn't dominate:
+LTX-2 mitigates both by inserting `mx.eval` at strategic points so each command buffer is small enough to fit one watchdog window but few enough that queue contention doesn't dominate. These guards apply on **all Apple Silicon Macs** — the previous 48 GB threshold was empirically wrong; M2 Max 64 GB machines exhibit `MTLCommandBufferErrorInternal` (code 14) crashes without them:
 
-- **Gemma forward**: per-layer eval (48 layers × ~100 ms each). Override via `LTX2_GEMMA_EVAL_EVERY=N` (default `1` on ≤48 GB, `0` on bigger Macs).
-- **TextEmbeddingProjection**: per-output-projection eval (splits the 188160→4096 video matmul from the 188160→2048 audio matmul). No-op on >48 GB.
-- **Embeddings1DConnector**: per-block eval (8 video + 8 audio blocks). No-op on >48 GB.
+- **Gemma forward**: per-layer eval (48 layers × ~100 ms each). Override via `LTX2_GEMMA_EVAL_EVERY=N` (default `1`).
+- **TextEmbeddingProjection**: per-output-projection eval (splits the 188160→4096 video matmul from the 188160→2048 audio matmul).
+- **Embeddings1DConnector**: per-block eval (8 video + 8 audio blocks).
+- **LTX DiT block loop**: eval every `LTX2_DIT_EVAL_EVERY` blocks (default `8`). Splits the 48-block forward into 6 command buffers of ~6 blocks each (~1–2 s/buffer), well within the watchdog window.
 
-`>48 GB` Macs (Mac Studio, M-series Ultra) keep full lazy-graph pipelining for max throughput.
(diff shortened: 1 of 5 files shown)


dgrauet:
(Auto-closed by #20 merge — adding the detailed explanation here as promised.)

Thanks again for the investigation @colinbdesign — v0.14.2 just shipped.

## Summary of what shipped vs. what was proposed here

### ✅ Fix 3 (LoRA silently ignored) — shipped, scope expanded
You correctly identified that `DistilledPipeline.load()` was dropping `_pending_loras`. Auditing more carefully, the same bug actually existed in `--one-stage`, `--two-stage`, and `--two-stages-hq` paths — every pipeline whose `load()` overrides `BasePipeline.load`. v0.14.2 fixes the wrapper level instead of patching each subclass: LoRA fusion now lives inside `_base.py::_load_transformer_with_optional_streaming`, so the four affected pipelines pick it up via the inheritance contract. The CHANGELOG credits you for surfacing the bug class.

A regression test (`tests/test_pending_loras_dispatch.py`) was added to lock the contract: future overrides that sidestep the wrapper will fail loudly.

### ❌ Fix 1 (`generate_from_image` on distilled) — not shipped
Two issues:
- The upstream commit reference `493aec2` doesn't exist on `Lightricks/LTX-2`. `DistilledPipeline` upstream is purely `__call__`-based (single method, no separate `generate_from_image`).
- No caller in our codebase checks `hasattr(pipe, \"generate_from_image\")`. Adding this method would diverge from upstream rather than mirror it.

If you're seeing tiling artefacts on I2V via `--distilled`, the upstream-iso pattern for static-scene identity preservation is `ic-lora` with a canny control video duplicated from the source image (documented in [CLAUDE.md → Static-scene I2V recipe](https://github.com/dgrauet/ltx-2-mlx/blob/main/CLAUDE.md#static-scene-i2v-recipe-preserve-identity)). The `--distilled` mode is explicitly not optimised for long-clip identity preservation — same as upstream.

### ❌ Fix 2 (extra watchdog guards in distilled.py) — not shipped
- The `LTX2_DIT_EVAL_EVERY` env var + block-loop eval guard you proposed already shipped via your earlier #3.
- The pre-materialisation + `min(2, default)` override + state-flush guards are net-new. Skipped from this cycle because (a) we couldn't validate them empirically on our hardware, and (b) adding them only to `distilled.py` creates asymmetry with the other two-stage pipelines that share the same code path. If you want to push these as a follow-up, a separate PR adding state-flush at the `_base.py` level (so all pipelines benefit) would be the right shape.

### Code quality notes (for any future PRs)
- The CLAUDE.md / Gemma encoder edits in this PR duplicate #3 — the branch is based on an older `main`. A rebase before re-opening would resolve the merge conflicts.
- Avoid inline imports + underscore-prefixed local aliases — they hurt readability.
- Module-level mutable state (`_dit_mod._DIT_EVAL_EVERY = ...`) with try/finally is fragile under concurrent generates. Prefer a kwarg threaded through the call.

Thanks again — the LoRA T2V bug class was a real find. Ping if you'd like to iterate on watchdog guards as a separate PR.

Issue #4, opened 2026-05-11
disfunction ...it was working befor teh update


Akossimon:
so after the updat it stopped working ....  i am getting a weird result , before the update it was alll fine , only cropping was not working correctly, but now all is disfunctiong. tell me waht you need,... terminal does not show any informations , which is odd

<img width="3824" height="2744" alt="Image" src="https://github.com/user-attachments/assets/7500c965-b58e-470b-b824-347231da50e8" />

dgrauet:
Hello, what is this app ?
Does it use ltx-2-mlx ?
Does it pin the version ?
Which version ?
Have you tried with the last fixes ?
Which pipeline has the issues ?

dgrauet:
I contacted the developer of this app so he can look at this issue, moreover this ticket should be created to the app repository https://github.com/mrbizarro/phosphene

Thank you for your feedback though, much appreciated

PR #25, opened 2026-05-15
(perf): VAE decode memory improvements

plz12345:
Note, this one was AI-assisted, after finding the current choke points while testing on a 32GB M5 Macbook Air.  VAE decode would throw it into heavy swap usage even though the rest of the pipeline stages never touched it.  I've been running with this code in my main pipeline flow (8/3 --distilled) for almost a week.  VAE memory only becomes an issue now if I make the resoution too high for the hardware (720p > 10 seconds of video).

### Problem

`VideoDecoder.stream_to_ffmpeg` decoded the full latent volume in a single
`self.decode()` call before streaming frames. At typical resolutions the
block-3 intermediate tensor (512 ch × 4× temporal × 4× spatial upsampling
at float32) can peak at 8–20+ GB for long or HD clips, causing OOM or heavy
swap before a single frame leaves the decoder.

The `tiled_decode` path also held on to intermediate activations across tiles
because neither the tile decode nor the accumulation buffers were evaluated
eagerly.

### Changes

**`_compute_decode_tiling` (new)** — pure arithmetic function that derives a
`TilingConfig` from the latent shape and an 8 GB peak-activation budget.
Estimates the block-3 bottleneck tensor size (`512 × 4F × 4H × 4W × float32`)
and returns `None` when the full video already fits, avoiding tiling overhead
for typical resolutions (≤720p, ≤~20s). When tiling is needed, tile size and
overlap are sized so overlap is approximately one second of output frames,
rounded to the nearest 8 frames (VAE temporal stride requirement), capped at
25% of tile size.

**`stream_to_ffmpeg`** — replaced the single `self.decode(latent)` + frame
loop with a `_compute_decode_tiling` call followed by
`self.tiled_decode(latent, tiling)`. At common resolutions `tiling` is `None`
and the code path is equivalent to before; at large resolutions tiling is
applied automatically without any caller change.

**`tiled_decode` — no-tiling path** — added `mx.eval` + `aggressive_cleanup`
after `self.decode` so block-3 intermediate activations are freed before the
caller begins streaming frames.

**`tiled_decode` — tiling path** — added `mx.eval` + `aggressive_cleanup`
immediately after each tile decode, and `mx.eval(buffer, weights)` after each
accumulation step. This breaks the lazy graph so the previous tile's
intermediate activations become evictable before the next tile decode begins.

### Tests (`tests/test_decode_tiling.py`)

17 pure-arithmetic tests (no weights, no GPU) covering `_compute_decode_tiling`:

- No tiling when video fits in budget; triggers one byte over
- `TemporalTilingConfig` constraints: tile size ≥ 16, divisible by 8,
  overlap < tile size, passes `__post_init__` validation
- Frame rate scaling: overlap grows monotonically with fps
  (24→24, 30→24, 48→48, 60→56 frames)
- Edge cases: minimum tile size enforced, zero overlap when tile is too small, default fps=24

### DIFF
--- a/packages/ltx-core-mlx/src/ltx_core_mlx/model/video_vae/video_vae.py
+++ b/packages/ltx-core-mlx/src/ltx_core_mlx/model/video_vae/video_vae.py
@@ -434,6 +434,13 @@ class VideoDecoder(nn.Module):
         """
         ffmpeg = find_ffmpeg()
         tiling = _compute_decode_tiling(latent.shape, frame_rate=frame_rate)
+        if tiling is not None and tiling.temporal_config is not None:
+            tc = tiling.temporal_config
+            logger.info(
+                "vae-decode tiled: tile_frames=%d overlap=%d",
+                tc.tile_size_in_frames,
+                tc.tile_overlap_in_frames,
+            )
 
         # Estimate output dimensions from latent
         _, _, _F_lat, H_lat, W_lat = latent.shape
--- a/packages/ltx-core-mlx/src/ltx_core_mlx/model/video_vae/video_vae.py
+++ b/packages/ltx-core-mlx/src/ltx_core_mlx/model/video_vae/video_vae.py
@@ -392,7 +392,9 @@ class VideoDecoder(nn.Module):
                 yield_len = curr_temporal_slice.start - previous_temporal_slice.start
                 if yield_len > 0:
                     safe_weights = mx.maximum(previous_weights, 1e-8)
-                    yield (previous_chunk / safe_weights)[:, :, :yield_len, :, :]
+                    chunk = (previous_chunk / safe_weights)[:, :, :yield_len, :, :]
+                    mx.eval(chunk)
+                    yield chunk
 
             previous_chunk = buffer
             previous_weights = weights
@@ -401,7 +403,9 @@ class VideoDecoder(nn.Module):
         # Yield remaining chunk
         if previous_chunk is not None and previous_weights is not None:
             safe_weights = mx.maximum(previous_weights, 1e-8)
-            yield previous_chunk / safe_weights
+            chunk = previous_chunk / safe_weights
+            mx.eval(chunk)
+            yield chunk
 
     def decode_and_stream(
         self,
@@ -466,16 +470,13 @@ class VideoDecoder(nn.Module):
         for chunk in self.tiled_decode(latent, tiling):  # (B, 3, T, H, W)
             if pipe_broken:
                 break
-            # chunk is already materialized by tiled_decode — async_eval is a no-op
-            # but kept for compatibility if the yield path changes.
-            mx.async_eval(chunk)
             num_frames = chunk.shape[2]
             for i in range(num_frames):
                 frame = chunk[:, :, i, :, :]
                 frame = mx.clip(frame, -1.0, 1.0)
                 frame = ((frame + 1.0) * 127.5).astype(mx.uint8)
                 frame_hwc = frame[0].transpose(1, 2, 0)  # (H, W, 3)
-                mx.eval(frame_hwc)  # sync before write — async_eval can race memoryview
+                mx.eval(frame_hwc)  # required: memoryview races GPU writes without this sync
                 try:
                     proc.stdin.write(bytes(memoryview(frame_hwc)))
                 except BrokenPipeError:
--- a/packages/ltx-core-mlx/src/ltx_core_mlx/model/video_vae/video_vae.py
+++ b/packages/ltx-core-mlx/src/ltx_core_mlx/model/video_vae/video_vae.py
@@ -56,6 +56,7 @@ from ltx_core_mlx.utils.memory import aggressive_cleanup
 
 logger: logging.Logger = logging.getLogger(__name__)
 
+
 def _compute_decode_tiling(
     latent_shape: tuple[int, ...],
     peak_budget_gb: float = 8.0,
@@ -69,7 +70,7 @@ def _compute_decode_tiling(
     _, _, F_lat, H_lat, W_lat = latent_shape
     budget_bytes = int(peak_budget_gb * 1024**3)
 
-    # Block-3 peak memory: 512 channels × 4 temporal frames × (4H spatial) × (4W spatial) × 4 bytes
+    # Block-3 peak memory: 512 channels x 4 temporal frames x (4H spatial) x (4W spatial) x 4 bytes
     block3_bytes_per_lat_frame = 512 * 4 * (H_lat * 4) * (W_lat * 4) * 4
 
     if block3_bytes_per_lat_frame * F_lat <= budget_bytes:
@@ -78,7 +79,7 @@ def _compute_decode_tiling(
     # How many latent frames fit within the budget?
     max_lat_frames = max(2, budget_bytes // block3_bytes_per_lat_frame)
 
-    # Convert latent frames → output pixel frames (8× temporal upsampling), with a 16-frame minimum
+    # Convert latent frames -> output pixel frames (8x temporal upsampling), with a 16-frame minimum
     tile_frames = max(16, max_lat_frames * 8)
 
     # Overlap ≈ 1 second of pixel frames at the given frame rate, rounded down to a multiple of 8,
@@ -296,8 +297,8 @@ class VideoDecoder(nn.Module):
         """
         if tiling_config is None:
             pixels = self.decode(latent)
-            mx.eval(pixels)      # materialize now — frees block-3 intermediate activations
-            aggressive_cleanup() # release GPU cache before caller begins streaming
+            mx.eval(pixels)  # materialize now — frees block-3 intermediate activations
+            aggressive_cleanup()  # release GPU cache before caller begins streaming
             yield pixels
             del pixels
             return
--- a/packages/ltx-core-mlx/src/ltx_core_mlx/model/video_vae/video_vae.py
+++ b/packages/ltx-core-mlx/src/ltx_core_mlx/model/video_vae/video_vae.py
@@ -64,14 +64,15 @@ def _compute_decode_tiling(
 ) -> TilingConfig | None:
     """Return a TilingConfig that keeps peak VAE decode memory under budget, or None.
 
-    Peak memory occurs at the block-3 intermediate tensor (the second DepthToSpaceUpsample), which has shape (1, 512, F_lat*4, H_lat*4, W_lat*4) at float32. We size tiles so thi tensor stays within budget. Returns None when the full video already fits — no tiling,
-    no overhead.
+    Peak memory occurs at the block-3 intermediate tensor (the second
+    DepthToSpaceUpsample), shape (1, 512, F_lat*4, H_lat*4, W_lat*4) in bf16.
+    Returns None when the full video fits within budget — no tiling, no overhead.
     """
     _, _, F_lat, H_lat, W_lat = latent_shape
     budget_bytes = int(peak_budget_gb * 1024**3)
 
-    # Block-3 peak memory: 512 channels x 4 temporal frames x (4H spatial) x (4W spatial) x 4 bytes
-    block3_bytes_per_lat_frame = 512 * 4 * (H_lat * 4) * (W_lat * 4) * 4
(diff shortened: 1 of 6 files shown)


dgrauet (commented):
Thanks for this — solid contribution and the problem statement is exactly right. Reviewed the diff in detail; a few items worth addressing before merge, plus some smaller polish.

## 🔴 Worth fixing before merge

**1. `except BrokenPipeError: pass` now wraps the tiled generator.** Previously this catch was a narrow guard around `proc.stdin.write`. After the rewrite it also wraps `self.tiled_decode(...)`, which runs N tile decodes. Any `BrokenPipeError` (or fd-related wrap) raised mid-tiling will silently return a truncated-but-valid-looking mp4. Suggest tightening the `try` to wrap only the `proc.stdin.write(...)` call, and/or `logger.warning(...)` with the frame count when caught.

**2. fp32 budget on a bf16 pipeline** (`video_vae.py:77`). The `× 4 bytes` is fp32; VAE runs in bf16, so real activation is half the estimated size — tiling kicks in at half the resolution it actually needs to. Either change the constant to `2` or document the 2× safety margin explicitly (the docstring claim "≤720p, ≤20s falls through" is calibrated against fp32 math, not the dtype actually used).

**3. Make the budget env-var driven.** Project convention is `LTX2_*_EVAL_EVERY` / `LTX2_GEMMA_MAX_LENGTH` for memory/perf knobs — 8 GB is reasonable on 32 GB Macs but punitive on Mac Studio 64/128 GB. Suggest `LTX2_VAE_DECODE_BUDGET_GB` with 8.0 default.

## 🟠 Smaller items

**4. `mx.async_eval(chunk)` is dead weight** (line ~457). `tiled_decode` already calls `mx.eval` before yielding, so this is provably a no-op today. More worryingly, the "kept for compatibility" framing invites a future reader to remove the load-bearing per-frame `mx.eval(frame_hwc)` below — which **would** race `memoryview` against GPU writes. Suggest deleting the `async_eval` and replacing the comment with a "do not remove the per-frame sync below — memoryview races GPU writes otherwise" note.

**5. `if overlap >= tile_frames: overlap = 0`.** Unreachable for valid inputs (overlap is upper-bounded by `tile_frames // 4`). If ever hit, silently zeroing produces visible seams. Replace with `assert overlap < tile_frames` so any future regression in the arithmetic fails loudly.

**6. Typo "thi tensor"** → "this tensor" in the new docstring.

**7. Phase marker.** When tiling kicks in, the numerical decode path changes silently. A `[phase] vae-decode tiled (N tiles, overlap=K)` marker via the existing phase-marker plumbing (`BasePipeline._decode_*`) would make this observable in CLI output.

## 🟡 Tests

The 17 arithmetic tests are great — fast, no GPU, well-scoped. The integration surface (the rewritten `tiled_decode` blend + the `decode_and_stream` chunk loop) is untested though. A bit-exact `tiled_decode(latent, cfg)` vs `decode(latent)` test on a tiny synthetic latent would close the highest-risk gap — `tests/test_modality_tiling.py` has the pattern (PSNR check on identity reconstruction). Happy to add this in a follow-up if you'd rather keep this PR focused.

## ✅ What's working well

- Clean separation: pure-arithmetic helper + thin integration. Easy to reason about, easy to revert.
- `None`-return short-circuit means zero overhead at common resolutions.
- Per-tile `mx.eval` + `aggressive_cleanup` aligns precisely with the CLAUDE.md "Streaming VAE Decode (NON-NEGOTIABLE)" + Metal-cache rules.
- A week of production use on your end is a real validation signal — appreciated.

Let me know if you'd like me to take any of the smaller items in a follow-up PR rather than asking you to spin again — happy to do that for #4–#7.

plz12345:
I'll review and come back.  I don't think we need to split it into separate PR's, but ack on your feedback.

plz12345:
@dgrauet I pushed these fixes as itemized commits to make re-review easier.  If you are good with them, I can squash merge for a clean PR commit.

dgrauet:
Thanks for the follow-up round — items 1-6 are clean, and 8 (integration test) got a test. Re-reviewed end-to-end against the final state. CI green and 34 tests pass locally in <1s. Two items still need work before merge, one new addition worth a quick measurement, and one doc nit.

## 🔴 Still blocking

### 1. The new tiling log is silent in CLI by default

`logger.info("vae-decode tiled: ...")` (commit 9, address-item-7) won't reach the user. Project's CLI doesn't call `logging.basicConfig`, and `phase()` (`utils/progress.py:19-35`) bypasses the `logging` module entirely — it does unconditional `print(..., file=sys.stderr)` gated on a `verbose` bool. So in a normal `ltx-2-mlx generate ...` invocation, this line emits nothing.

Cleanest fix: emit from the pipelines-side wrapper, not the core helper. `VideoDecoder.decode_and_stream` in `packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/utils/blocks.py` already has access to `self.verbose` and is the right layer for user-visible markers. Have it call `_compute_decode_tiling(...)` once before delegating, and print on stderr if tiling kicks in:

```python
tiling = _compute_decode_tiling(latent.shape, frame_rate=frame_rate)
if tiling is not None and tiling.temporal_config is not None and self.verbose:
    tc = tiling.temporal_config
    print(f"[vae-decode tiled: tile_frames={tc.tile_size_in_frames} overlap={tc.tile_overlap_in_frames}]", file=sys.stderr, flush=True)
```

You can then pass the precomputed `tiling` into the core helper as a kwarg (avoids computing twice) or just leave the core helper to recompute — it's pure arithmetic, free.

### 2. The integration test is degenerate

`test_tiled_decode_single_tile_matches_decode` passes a config (`tile_size_in_frames=32`) larger than the full latent (F_lat=2 → 16 pixel frames), so `prepare_tiles_for_decoding` returns **one** tile. The for-loop over `temporal_groups` runs exactly once, the `if previous_chunk is not None` branch is never reached, and the blend math you actually rewrote isn't exercised. The test reduces to `(decoded × 1.0) / 1.0 == decoded`.

The risk surface this PR introduces — the per-tile materialization, the accumulation buffers, the overlap blend — is all on the **multi-tile** path. A second test would close that:

```python
def test_tiled_decode_multi_tile_matches_decode(self):
    """Tiled decode with >=2 tiles + overlap must reconstruct the baseline."""
    mx.random.seed(0)
    decoder = VideoDecoder()
    mx.eval(decoder.parameters())
    latent = mx.random.normal((1, 128, 5, 3, 3))  # 33 pixel frames -> at least 2 tiles
    mx.eval(latent)
    baseline = decoder.decode(latent); mx.eval(baseline)
    cfg = TilingConfig(temporal_config=TemporalTilingConfig(
        tile_size_in_frames=16, tile_overlap_in_frames=8,
    ))
    chunks = list(decoder.tiled_decode(latent, cfg))
    tiled = mx.concatenate(chunks, axis=2); mx.eval(tiled)
    assert tiled.shape == baseline.shape
    # Tiny atol since accumulation order differs; this is the bound we hit on
    # tests/test_modality_tiling.py for the equivalent invariant.
    assert mx.allclose(baseline, tiled, atol=1e-4, rtol=1e-4).item()
```

If atol needs widening that's informative too — it tells us the blend math is not bit-exact and we should document the precision floor.

## 🟡 New in this round, not yet measured

### 3. Commit 11 (per-stage materialization in `decode`)

Not part of the original review — appeared in commit 11. It adds 4 sync points per `decode()` call, breaking kernel fusion across the 4 upsample stages. In the tiled path that's `4 × N_tiles` extra syncs, on top of the per-tile sync already added in commit 2. Memory benefit is real (frees pre-upsample activation before the next, larger one), but the perf cost is unmeasured.

Could you share a single timed run of `ltx-2-mlx generate --distilled` at your usual config with and without that hunk, just to confirm it's not regressing the common no-tiling case? On the no-tiling path, `tiled_decode` already materializes pixels before yielding, so the per-stage syncs inside `decode()` are partial duplicates of work that was going to happen anyway — but the bookkeeping cost of 4 extra sync points isn't zero.

If the run shows >5% wall-clock regression on the no-tiling path, gating commit 11 on a flag passed into `decode()` (only set true from the tiled path) would keep the win where it matters.

## 🟢 Doc nit

The `decode_and_stream` docstring at line 422-430 says "exceed the memory budget" measured against block-3. Accurate for the *peak inside one tile decode*, but the accumulation `buffer` and `weights` are allocated at fp32 default — at 1080p × 200 frames that's another ~10 GB on top of the per-tile budget. Worth a one-liner:

> Budget covers the block-3 activation of a single tile decode. The accumulation buffer (fp32 `B × 3 × T × H × W` and its weights twin) lives on top of this; at 1080p / >100 frames it adds several GB.

Or switch the buffer to bf16 — the math is already lossy by virtue of trapezoidal blending, an extra bf16 round-trip on the accumulator likely doesn't change the qualitative output. Up to you, this one isn't blocking.

---

Sorry for the second round — items 1 and 2 are quick. Once those land I'm happy to merge. The week of in-production validation on your end is solid evidence for the core claim; the asks here are about making sure (a) other users actually see when it kicks in and (b) the regression net catches the blend math, not just the dispatch.

plz12345:
No worries, iteration and healthy reviews are never a bad thing.  I've pushed 4 commits targeting your recent round.  Regarding item 3, here are the results of a 48p 15s 25FPS run, with and without the mx.eval call.  Since that did breach your note on 5%, I reworked it to only fire during a tiling job.  That is in place, so we skip that mx.eval unless the job is tiled.  The other changes in the PR already do a solid job if decreasing the decode stage resource hits, and the 32 -> 16 one (upcoming PR, I want more testing) will take that further.

Without
```
Denoising: 100%|██████████| 8/8 [03:10<00:00, 23.81s/it]
Denoising: 100%|██████████| 3/3 [08:26<00:00, 168.79s/it]
[Loading decoders (VAE + audio + vocoder)] ...
[Loading decoders (VAE + audio + vocoder)] done in 0.0s
[Decoding video + audio + muxing] ...
[Decoding video + audio + muxing] done in 124.1s

Saved to: ~/Downloads/ltx_output/clip_20260520_001705_3886407087.mp4
Time: 841.0s
```

With
```
Denoising: 100%|██████████| 8/8 [03:27<00:00, 25.89s/it]
Denoising: 100%|██████████| 3/3 [08:43<00:00, 174.45s/it]
[Loading decoders (VAE + audio + vocoder)] ...
[Loading decoders (VAE + audio + vocoder)] done in 0.1s
[Decoding video + audio + muxing] ...
[Decoding video + audio + muxing] done in 131.4s

Saved to: ~/Downloads/ltx_output/clip_20260519_235901_2859475233.mp4
Time: 881.9s
````

| | With `mx.eval` | Without | Delta | Delta % |
|---|---|---|---|---|
| Denoise stage 1 | 207s | 190s | +17s | +8.9% |
| Denoise stage 2 | 523s | 506s | +17s | +3.4% |
| Decode | 131.4s | 124.1s | +7.3s | +5.9% |
| **Total** | **881.9s** | **841.0s** | **+40.9s** | **+4.9%** |

dgrauet (approved):
All four items addressed cleanly:

1. **silent log** (commit f5d3fc38) — `verbose` wired through `VideoDecoder` in `utils/blocks.py`, emits via `print(..., file=sys.stderr)` from the wrapper layer. Exactly the right seam.
2. **degenerate test → multi-tile test** (commit 01bd0d94) — and the *reason* my proposed `mx.allclose(baseline, tiled)` doesn't hold is itself a real finding: causal `Conv3dBlock` padding (`convolution.py:62-66`) replicates each tile's first frame, so isolated tiles diverge from the full-decode context. Trapezoidal blend smooths visually but doesn't reconstruct the exact signal. Worth a brief mention in the CHANGELOG when this lands — users dropping into tiled mode at 1080p+ should know the output isn't bit-for-bit the same as the same-config-with-more-RAM no-tile run.
3. **commit 11 perf** (commit b208e3d8) — measured +5.9% on decode, gated to tiled-only via `_materialize_stages: bool = False` kwarg on `decode()`. The no-tiling fast path keeps kernel fusion intact.
4. **doc nit** (commit 4298b81c) — accumulation-buffer caveat documented, fp32→bf16 deferred to a separate PR with a clear TODO. Fine — keeps this PR focused.

Bonus a2v consistency cleanup (commit 8b2887b8) is welcome — `assert self.vae_decoder is not None` was a smell, and routing through `video_decoder_block` matches the rest of the plumbing.

35/35 tests pass locally in 2.4s. CI green. Ready to squash-merge.

Issue #6, opened 2026-05-13
REQ: Support for FPS in Generate pipelines


plz12345:
LTX 2.3 docs say FPS is settable, and they even do it in their ComfyUI nodes.  Requesting ability to do it in the ltx-2-mlx pipeline as well.  If this needs to be requested upstream, can you point me in the right direction?

Issue #5, opened 2026-05-12
REQ: CLI output for silent phases in pipeline


plz12345:
There are some phases that can be high load, or long running, that are silent during the pipeline as there are current only CLI outputs for the pipeline type, and the two stages.  Requesting CLI outputs for:

1. Text encoder.  I know this is pretty quick, but it's invisible to the user and important to the process.
2. Loading transformer, the dev models are big and depending on the machine, this can take a little bit.
3. VAE decode, post stage 2, this is another point that is hidden by a heavy demand on the system.  The user would appreciate knowing why.

I'm aware these likely all print to stderr in the native Python code, but I don't think "write python" is a practical solution in this case.  Thanks, I hope, in advance.

dgrauet:
Hello, could you give some feedback with the last patch ?

dgrauet:
Done — shipped in [**v0.13.1**](https://github.com/dgrauet/ltx-2-mlx/releases/tag/v0.13.1).

Each pipeline now prints bracketed `[phase] ...` / `[phase] done in X.Ys`
markers on **stderr** around the five long silent stages:

1. Loading text encoder (Gemma)
2. Encoding prompt
3. Loading transformer
4. Loading decoders (VAE + audio + vocoder)
5. Decoding video + audio + muxing

Output goes to stderr so stdout stays clean for callers piping pipeline
output. Suppressed by `--quiet`. Thanks for the report.

Issue #2, opened 2026-05-08
REQ: Add ability to supress audio generation/inclusion


plz12345:
Unless I missed something, this isn't currently supported, so the generations have a habit of getting unexpected music/effects.

dgrauet:
There is currently nothing even upstream to suppress the audio. For now, the audio can be removed easily with ffmpeg but I understand that this is one more step and technically more compute time for the generation.

plz12345:
Yeah, it's annoying as hell.  Random I2V clips will just have background music for no reason.  I'm playing around with treating silence as music, to override it.

PR #30, opened 2026-05-20
feat: enable lora support on low RAM flow, minor tweaks

plz12345:
This builds off of PR #20

After that PR, we still had some gaps where LoRAs couldn't be used, namely, in combination with --low-ram .  This PR does the following:

- The `BlockLoraSource` streaming path — removes the `NotImplementedError` and makes `--lora --low-ram` work on all pipeline variants
- Corrects the cli.py help text to reflect the new reality
- Simplifies `_remap_lora_keys` in ti2vid_two_stages.py (uses the shared map instead of hardcoded audio key patches)
- Extracts `resolve_lora_path` into `_orchestration.py` so both the low-ran and normal paths share the same HuggingFace download logic

All tests and lint pass locally, and running generate --low-ram --lora /path/to/lora.safetensors also runs cleanly with multiple LoRAs fused at 1.0 and custom strengths.

I'm sure this may require an addiotional passthourgh, no worries.


dgrauet:
## Review — `feat: enable lora support on low RAM flow`

Thanks for the follow-up on #20! Three review passes (general / tests / silent-failures) + manual verification of the `apply_to_key` contract.

### Verdict: **LGTM with minor polish suggestions** — no blockers.

The architecture is sound. The `--low-ram + --lora` wiring in `_base.py:283-301` mirrors `ICLoraPipeline._fuse_loras` verbatim (same `BlockLoraSource` ctor, same `block_prefix`, same `object.__getattribute__/__setattr__` on `_lora_sources` — that's the established `StreamingLTXModel` convention, not a workaround).

### Suggestions (none blocking)

**1. Tighten the new streaming test** — `tests/test_pending_loras_dispatch.py`
`BlockLoraSource_cls.assert_called_once()` doesn't check args. A regression that passes the wrong `block_prefix` (e.g. drops the `\"transformer.\"` outer prefix → zero blocks match → silent no-op LoRA) wouldn't be caught. Replace with:
\`\`\`python
BlockLoraSource_cls.assert_called_once_with(
    \"/fake/lora.safetensors\",
    block_prefix=\"transformer.transformer_blocks.\",
    strength=1.0,
    sd_ops=LTXV_LORA_COMFY_RENAMING_MAP,
)
\`\`\`
And add a parametrized multi-LoRA case (\`[(p1, 0.8), (p2, 1.2)]\` → \`call_count == 2\`) — the PR description claims this works but it's untested.

**2. Add `resolve_lora_path` unit tests** — `utils/_orchestration.py:39-60`
4 branches, zero coverage: local-exists, HF-single, HF-multi-warn, HF-zero-raises. Trivial with `tmp_path` + monkeypatched `snapshot_download`.

**3. Multi-safetensors picks alphabetical-first silently** — `utils/_orchestration.py:58-60`
`safetensors_files[0]` is filesystem-order dependent; user gets a different LoRA than they asked for with only a `_logger.warning`. Two options: (a) raise — \`ValueError(f\"Ambiguous: {len(files)} .safetensors in repo {path}; specify which one\")\` — safer; (b) keep current behavior but at least \`sorted(...)\` for determinism. Recommend (a).

**4. Logger output likely swallowed** — `utils/_orchestration.py:50,52,59`
`logging.getLogger(__name__).info/warning` with no `basicConfig` configured anywhere → INFO drops. Two lines later \`fuse_pending_loras:80\` still does \`print(f\"  Fusing LoRA: ...\")\`. The rest of the codebase uses \`phase()\` markers on stderr or bare \`print\`. Either switch to \`print(..., file=sys.stderr)\` for consistency, or merge with #3 (raise on multi → no warning needed).

**5. `_remap_lora_keys` `or k` fallback is dead code** — `ti2vid_two_stages.py:87`
Verified: \`LTXV_LORA_COMFY_RENAMING_MAP\` uses \`with_matching()\` (empty prefix/suffix matches any key) and no \`allowed_keys\`, so \`apply_to_key\` always returns a string on this map (\`sd_ops.py:94-110\`). The \`or k\` is harmless but misleading. Either drop it or add a one-line comment that it's defensive against future \`allowed_keys\` additions.

**6. Test the shared-map move** — `sd_ops.py:140-146`
The 4 new replacements (\`.linear_1.\`, \`.linear_2.\`, \`audio_ff.net.0.proj.\`, \`audio_ff.net.2.\`) are now load-bearing for the streaming path (which uses the map directly, no longer goes through \`_remap_lora_keys\`). A 4-row parametrize asserting \`LTXV_LORA_COMFY_RENAMING_MAP.apply_to_key(\"...linear_1...\") == \"...linear1...\"\` etc. locks the contract.

### Strengths

- Correct IC-LoRA pattern reuse — no architectural drift; \`BlockLoraSource\` args match verbatim.
- Shared-map consolidation in the right place. New replacements are no-ops on keys that don't contain the source substring, so IC-LoRA / two-stage paths are unaffected.
- CLAUDE.md and CLI help text updated consistently with the behavior change.
- Test updated alongside (old \`raises(NotImplementedError)\` removed, no stale negative assertion).
- Clean, focused, single-concern PR.

### Suggested action plan

1. (recommend) Tighten the test as in #1.
2. (recommend) Decide between #3-raise or #3-sort + #4-print.
3. (nice) Add \`resolve_lora_path\` tests (#2) and shared-map test (#6).
4. (optional) Strip \`or k\` in #5.

🤖 Review co-authored with Claude Opus 4.7

plz12345:
All reasonable, I'll work on that and update the PR.

PR #32, opened 2026-05-31
feat: versioned safetensors loading

plz12345:
LTX2.3 Model dirs contain versioned filenames (e.g.
transformer-distilled-1.1.safetensors). This PR makes the loader
handle them correctly end-to-end:

- Dynamic resolution for LoRAs and transformers — picks the
  alphabetically latest versioned file when present, no hardcoded names
- Upscaler file naming and key prefix fixes
- Versioned safetensors now win over unversioned when both exist in the
  same model dir (previously the unversioned exact name took priority,
  silently loading the older file)
- If a user only had the newer file, to safe disk space, the pipeline would fail hard due to the hardcoded naming
- Weight loader falls back to safe_open for extensionless HF cache
  blobs, which mx.load can't parse without a known extension

### DIFF
--- a/packages/ltx-core-mlx/src/ltx_core_mlx/loader/sft_loader.py
+++ b/packages/ltx-core-mlx/src/ltx_core_mlx/loader/sft_loader.py
@@ -6,12 +6,22 @@ Ported from ltx-core/src/ltx_core/loader/sft_loader.py
 from __future__ import annotations
 
 import json
+from pathlib import Path
 
 import mlx.core as mx
 from safetensors import safe_open
 
 from ltx_core_mlx.loader.primitives import StateDict
 from ltx_core_mlx.loader.sd_ops import SDOps
+
+_KNOWN_MLX_EXTENSIONS = {".safetensors", ".npz", ".npy", ".gguf"}
+
+
+def _load_weights(path: str) -> dict[str, mx.array]:
+    """Load weights, handling extensionless HF cache blobs via explicit format."""
+    if Path(path).suffix in _KNOWN_MLX_EXTENSIONS:
+        return mx.load(path)
+    return mx.load(path, format="safetensors")
 
 
 class SafetensorsStateDictLoader:
@@ -43,7 +53,7 @@ class SafetensorsStateDictLoader:
 
         model_paths = path if isinstance(path, list) else [path]
         for shard_path in model_paths:
-            weights = mx.load(shard_path)
+            weights = _load_weights(shard_path)
             for name, value in weights.items():
                 expected_name = name if sd_ops is None else sd_ops.apply_to_key(name)
                 if expected_name is None:
--- a/packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/_base.py
+++ b/packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/_base.py
@@ -258,6 +258,21 @@ class BasePipeline:
             )
         return self._load_transformer_with_optional_streaming(dev_path)
 
+    @staticmethod
+    def _resolve_safetensors(model_dir: Path, stem: str) -> Path:
+        """Return the path for a (possibly versioned) safetensors file.
+
+        Tries ``{stem}.safetensors`` first (exact match), then globs for
+        ``{stem}-*.safetensors`` (e.g. ``transformer-distilled-1.1.safetensors``).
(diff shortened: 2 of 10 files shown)


dgrauet:
Thanks for this — the versioned resolution (`_resolve_safetensors`) is clean and the silent "unversioned wins" bug fix is a real one.

One question on the `_load_weights` safe_open fallback in `sft_loader.py`, before I can sign off on it:

**Have you actually hit an extensionless blob in practice, and through which path?** I traced every caller of `SafetensorsStateDictLoader.load` in the current code and they all resolve through `resolve_lora_path()` / `snapshot_download()`, which always return paths ending in `.safetensors` (HF returns snapshot symlinks with real filenames, not `blobs/<sha>`). So the `else` branch never seems to be reached in the normal flow — unless you're passing a raw `blobs/<sha>` path to `--lora` directly?

This matters because the fallback as written doesn't work on our weights:

```python
with safe_open(path, framework="numpy") as f:
    return {k: mx.array(f.get_tensor(k)) for k in f}
```

- `for k in f` raises `TypeError: 'builtins.safe_open' object is not iterable` (safetensors 0.7.0 has no `__iter__`; needs `f.keys()`) — fails on *any* blob, even f32.
- Even with `.keys()`, `framework="numpy"` can't decode bf16: `TypeError: data type 'bfloat16' not understood`. Our transformer weights are bf16/int8.

If you do need the fallback, MLX handles extensionless safetensors directly with an explicit format (verified on bf16):

```python
def _load_weights(path: str) -> dict[str, mx.array]:
    if Path(path).suffix in _KNOWN_MLX_EXTENSIONS:
        return mx.load(path)
    return mx.load(path, format="safetensors")
```

If you 're not actually hitting this path, simplest is to drop the fallback entirely.

Separately: the `--distilled` CLI change (`image=` → `images=`) looks unrelated to versioned loading and removes the `_legacy_single_image` validation guard — could that move to its own PR so this one stays scoped to the loader change?

plz12345:
safe_open - yes, it happens if you point it at a HF hub snapshots GUID folder where the symlinked filenames live.  When it follows the symlink, the code loads the GUID blob by its real file name, ignoring the symlink's filename, leading to a throw.  It is an edge case in that it would only impact someone using a different model than one of yours (refinement, different distillation, custom learning version, etc.).  I'll take a look at your notes to see where I went astray, but the dynamic loading is a solid improvement because the current code ignores LTX's newer file versions at present, and I wanted to future proof that so all you need to do is upstream sync and re-forge your models, to have the code leverage those.

Regarding the image change on distilled, yeah, I can peel that off.  Sorry about that.

plz12345:
updated per your edit recommendation and the CLI image bit will be a separate PR.

dgrauet:
Both changes look good — thanks for the quick turnaround.

- `_load_weights` now uses `mx.load(path, format="safetensors")`. Verified it loads an extensionless bf16+uint32 blob correctly (and leaves the normal `.safetensors` path untouched). And thanks for the clarification on reachability — you're right, my trace was incomplete: pointing at a `snapshots/` dir that resolves through to the real GUID blob name does hit it. Real edge case for custom/refined models, good to have it work.
- CLI `image=`/`images=` change peeled off — appreciated, no need to apologize. Happy to review that one on its own.

Two non-blocking nits, your call:
- Version selection is alphabetical (`sorted(...)[-1]`), so `-1.10` would sort *below* `-1.9`. Fine for the current `1.0`/`1.1` scheme — just a latent footgun worth a comment in the docstring if LTX ever hits double-digit minors.
- Now that we know the extensionless path is genuinely reachable, a small unit test for `_load_weights` (extensionless blob → loads) would lock it against regressions, since that branch is invisible to the existing tests.

LGTM to merge once you're happy with those — the versioned resolution is a genuine improvement.

dgrauet:
Pushed the two nits directly to the branch (maintainer edit, `fcaad27`):
- Two unit tests in `tests/test_loader.py` for `_load_weights` — extensionless blob (bf16+uint32, the case you described) and the normal `.safetensors` path.
- A docstring note in `_resolve_safetensors` flagging the lexicographic version-sort limit (`-1.10` < `-1.9`).

Both green locally (`pytest tests/test_loader.py` → 26 passed, ruff clean). Feel free to squash/reword. From my side this is good to merge once CI is green.

plz12345:
You had the commit pushed before I even finished reading the first comment email, hehe, good on my end.

PR #41, opened 2026-06-10
fix: run VAE decoder decode in float32 for PyTorch parity

JanGrohn:
bf16 decode diverged from the upstream fp32 reference (up to ~12% on keyframe-sharp latents). Promote weights to float32 on first decode and compute the forward pass in fp32.

VAE decoder weights use ~1.6 GB fp32 during decode vs ~814 MB bf16 on disk. Acceptable in practice because DiT is freed before decode in all pipelines.


dgrauet (changes requested):
Thanks for digging into this and bringing real numbers — a ~12% delta on keyframe-sharp latents is worth taking seriously, and the dtype-promotion-once approach is clean.

That said, I can't take this as a parity fix, because **upstream doesn't decode the VAE in fp32** — I checked both reference paths:

- **Native ltx-core** (`packages/ltx-core/.../video_vae.py`): `VideoDecoder.forward` does `weights_dtype = next(self.parameters()).dtype; sample = sample.to(weights_dtype)` and the pipelines load it with `self.dtype = torch.bfloat16`. There's no `.float()` / `torch.float32` / autocast on the decode path, and the VAE configurator has no `force_upcast`-style flag.
- **Diffusers `AutoencoderKLLTX2`**: `decode()` runs in `vae.dtype` (bf16) with no `force_upcast`; only the transformer's `noise_pred` is upcast for guidance, not the VAE.

Our current `decode()` (the `b604d3f` dtype-preserving guard) is already byte-for-byte equivalent to upstream's bf16 path, so forcing fp32 here actually *diverges* from parity rather than restoring it. We keep this port strictly iso with ltx-core, so I'd rather not introduce a deliberate divergence framed as parity.

The ~12% you measured is almost certainly real — bf16 is genuinely less precise than fp32 — but "the fp32 reference" isn't what the upstream pipeline produces. If we want to chase decode quality, the right shape is a *targeted* higher-precision fix (identify the specific op that loses precision — likely a norm or the final unpatchify — and upcast just that), not a whole-decoder fp32 promotion that doubles decoder memory permanently. Happy to look at op-by-op MLX-vs-PyTorch diffs with you if you're up for it.

Two more asks before this could land in any form:

- **Please split out the `-shortest` ffmpeg change** — it's an unrelated (and plausibly valid!) extend/retake truncation fix and deserves its own PR with its own test. We keep one concern per PR so each carries a coherent version bump.
- A regression test would be needed for whatever the final fix is.

Minor note on the code itself: `_ensure_compute_dtype` only checks `flat[0]` for the dtype, which works today only because `conv_in` sorts first — it'd silently skip promotion if params were ever mixed-dtype.

Really appreciate the contribution — the measurement is the valuable part here, and I'd love to land a targeted version of it.

dgrauet (changes requested):
Thanks for digging into this, Jan — the keyframe pulsing is a real annoyance and I appreciate the investigation. I dug into the upstream reference before merging, though, and I don't think we should take the fp32 decode change. Let me lay out why.

**Upstream actually runs VAE decode in bf16, not fp32.** In `ltx-core`, `VideoDecoder.decode` runs the forward in the *weights* dtype:

```python
# ltx-core video_vae.py (~L702)
output_dtype = sample.dtype
weights_dtype = next(self.parameters()).dtype
sample = sample.to(weights_dtype)   # decode runs in weights dtype
...
return sample.to(output_dtype)
```

and every pipeline loads the VAE in bf16 — `self.dtype = torch.bfloat16`, including `keyframe_interpolation.py`. There's no VAE-specific fp32 override anywhere upstream.

That means our existing code (`cast latent → weights_dtype`) was already a line-for-line match of upstream's behavior. The ~12% gap you measured is a genuine bf16-vs-fp32 numeric difference, but it's against a PyTorch VAE loaded in fp32 — *not* against upstream's actual bf16 production decode path, which is the reference this port tracks. So this PR would make us more precise than upstream, at the cost of diverging from it, plus a RAM hit (~+800 MB of persistent fp32 weights and ~2× activation memory during decode, which bites on `--low-ram` HD and works against modality tiling).

This repo's standing policy is to stay iso with upstream unless there's a deliberate, documented reason to diverge. From earlier investigation the keyframe pulsing looks model-level (upstream bf16 exhibits it too), so fp32 here would be masking an upstream artifact rather than fixing a port bug. If you've got a before/after keyframe clip where fp32 visibly removes something bf16 can't, I'm open to revisiting it — but as an explicit opt-in/documented divergence, not framed as parity.

**On the `-shortest` removal:** that's an unrelated change (and not mentioned in the commit message), so it should go in its own PR. One thing to think through there: without `-shortest`, ffmpeg encodes to the *longest* input, so if the reconstructed audio is *longer* than the video you'd get extra duration with a frozen final frame — the inverse of the truncation you're fixing. Worth handling both directions explicitly.

Happy to take the audio fix separately if you want to split it out. Thanks again!

dgrauet:
Thanks for this, @JanGrohn — closing in favour of #58, which cherry-picks the parts that match the LTX reference.

**Merged (via #58):** the `-shortest` removal. Verified against upstream `ltx_pipelines/utils/media_io.py`, which muxes both streams in full (all video chunks + the entire audio waveform, no truncation). Our `-shortest` deviated from that and truncated tail frames on extend/retake. Your fix is correct and is now on `main`.

**Not taken:** the fp32 VAE decode. Checked against `ltx_core/model/video_vae/video_vae.py::forward` — upstream decodes the video VAE in the **weights dtype (bf16)** (`sample.to(weights_dtype)`), with no `autocast(float32)` / `.float()`; the pipeline even labels it "bfloat16 VAE". The fp32 autocast upstream is confined to the **audio** vocoder, not the video decoder. So forcing fp32 here would deviate from the reference. If MLX bf16 decode really diverges ~12% from a PyTorch bf16 reference, that points to a numerical/accumulation issue we'd rather diagnose directly than paper over with an fp32 promotion — happy to dig into that separately if you can share the latent/repro that showed the 12%.

Credit for the merged change is preserved in the #58 commit body.

PR #43, opened 2026-06-13
feat: audio LoRA training — preprocessing, slicer, and gradient checkpointing

plz12345:
### Summary

Enables end-to-end joint audio-video LoRA training on Apple Silicon, closing the gap between what the trainer config accepted and what it could actually execute. Three logical pieces:

- **Bug fixes + gradient checkpointing** — four latent bugs in the existing trainer that would crash any audio training run (`fps=` keyword in three call sites, decoders called as functions instead of `.decode()`, bfloat16 arrays passed directly to numpy). Gradient checkpointing on `LTXModel` caps backprop activation memory at ~1 block (~26 GB) vs. storing all 48 (~84 GB), making the dev model trainable on 64 GB machines.

- **Audio preprocessing pipeline** — `preprocess --with-audio` encodes each clip's audio track via the audio VAE encoder and writes `audio_latents/` alongside video latents, sized to `compute_audio_token_count()` so the two modalities are aligned by construction. Preprocessor now does a partial HuggingFace download (connector + VAE files only, skipping the ~20 GB transformer) and discovers clips recursively so per-source subfolders work. Adds `transformer_file` config field to pin `transformer-dev.safetensors` explicitly, and expands `~` in `model_path` validation.

- **Video slicer** — new `ltx-2-mlx slice` command cuts long source videos into normalized training clips with audio retained: fixed-interval or timecode-list slicing, aspect-safe crop/pad, `--max-clips` with even or sequential sampling, per-source output subfolders.

### Behavior change: preprocess no longer downloads the full model snapshot

Previously, `preprocess` with a HuggingFace repo ID as `--model` would trigger a full `snapshot_download`, pulling the entire model (~80 GB including the ~20 GB transformer). This PR changes `_resolve_model_dir` to do a **partial download**: only the connector and VAE safetensors files needed for preprocessing are fetched. The transformer is not downloaded.

**Impact**: anyone who relied on `preprocess` to populate the full HF cache (e.g. to then run inference from the same path) will need to run `huggingface-cli download <repo>` separately, or pass the already-cached local path. The changelog entry for this release should note the change.

### Validated

Trained a 2000-step audio-style LoRA on M5 Pro 64 GB Macbook Pro (74 clips, 8 voices, 192×192, rank 32) targeting `audio_attn1 / audio_attn2 / video_to_audio_attn` only. Confirmed at inference via `generate --lora`:
- Soft-spoken and whisper registers both work easily for female voices
- Male soft-spoken works easily; male whisper responds to `(man, whispering quietly)` (needed more of a nudge, likely due to source clips
- Video output visually unchanged from base model (audio-only target modules hold)
- ~5.7 s/step with gradient checkpointing at 192×192; ~3h10m for 2000 steps vs OOM without gradient checkpointing

### Test plan

- [x] `TestGradientCheckpointing::test_grads_match_on_off` — verifies `mx.checkpoint` with explicit param input gives identical LoRA-param grads to non-checkpoint path (catches silent no-op adapter regression)
- [x] `TestFitAudioTokens` (5 tests) — verifies `_fit_audio_tokens` trims/pads to the exact count from `compute_audio_token_count` for canonical (num_frames, fps) values
- [ ] `ltx-2-mlx slice <source.mp4> --interval 4 --res 384x384 -o ./clips` — confirm per-source subfolder created, clips have audio stream
- [ ] `ltx-2-mlx preprocess --videos ./clips --with-audio --frame-rate 24 -o ./data` — confirm `audio_latents/` written with matching filenames and count equal to `latents/`
- [ ] `ltx-2-mlx train --config configs/lora_av_whisper.yaml` — confirm training starts, loss is finite, checkpoints written
- [ ] `ltx-2-mlx train --config configs/lora_av_whisper.yaml --low-ram` — confirm gradient checkpointing enabled message, peak RAM lower than without flag
- [ ] `ltx-2-mlx generate --lora <checkpoint> --prompt "(woman, whispering) ..."` — confirm audio register effect at inference

### DIFF
--- a/packages/ltx-core-mlx/src/ltx_core_mlx/model/transformer/model.py
+++ b/packages/ltx-core-mlx/src/ltx_core_mlx/model/transformer/model.py
@@ -219,6 +219,11 @@ class LTXModel(nn.Module):
             for _ in range(config.num_layers)
         ]
 
+        # Training-only: recompute each block in the backward pass instead of
+        # storing all 48 blocks' activations. Caps activation memory at ~1 block
+        # so backprop through the dev model fits on 64 GB. No effect on inference.
+        self.gradient_checkpointing = False
+
     def _embed_timestep_scalar(
         self,
         timestep: mx.array,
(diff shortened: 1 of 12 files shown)


dgrauet (commented):
**Review — approve with two small test requests**

Thanks for this — it's a thorough, well-documented PR, and the bug fixes alone justify it. Validated everything against current `main`.

**The bug fixes are real and well-targeted.** All four are genuine crashers, not cosmetic:
- The `fps=` → `frame_rate=` changes in `trainer.py` / `base_strategy.py` / `validation_sampler.py` are regressions from the v0.14.0 iso-strict rename that missed the trainer package — same class of miss as the v0.14.1 decode-wrapper bug. Confirmed the core signatures are `frame_rate=`-only now, so any audio training run would have crashed on these.
- The `_vae_decoder(latent)` / `_audio_decoder(latent)` → `.decode(latent)` fixes are correct — the decoder classes expose `.decode()` and have no top-level `__call__`.
- The bf16→f32 cast before `np.array(...)` is right; numpy has no bfloat16 buffer dtype.

**Gradient checkpointing is implemented correctly**, including the subtle part: passing `block.trainable_parameters()` as an explicit `mx.checkpoint` input and rebinding via `block.update(params)` inside the closure. Without that, `mx.checkpoint` (which only tracks gradients w.r.t. its inputs) would leave the LoRA params at init — a silent no-op adapter. The default-arg loop-var capture (`_block`, `_bidx`) correctly avoids the late-binding closure trap. Default `False` + `continue` means zero inference impact, and it mirrors upstream's `gradient_checkpointing` on the transformer.

**Two requests before merge:**

1. **Add minimal test coverage.** The empirical 2000-step run is great validation, but CI can't re-run it. Two cheap unit tests would lock in the riskiest logic:
   - Gradient-checkpointing on/off equivalence: one block, compare grads w.r.t. LoRA params with `mx.allclose`. This is the piece most likely to silently regress.
   - Audio-token alignment: `_fit_audio_tokens` trim/pad to `compute_audio_token_count(...)`.

2. **Flag the preprocess behavior change in the description.** `_resolve_model_dir` now does a partial HF download (encoder files only) for repo-ID inputs instead of the full snapshot. It's a clear win (~80 GB saved), but it's a silent behavior change for anyone who relied on preprocess populating the full cache — worth a line so it lands in the changelog.

**Minor / non-blocking:** the example YAML has a `keep_last_n: 25  # keep all 20` mismatch, and the memory figures drift a bit between the config comment (~15-20 GB) and the PR body (~26 GB). Cosmetic.

This is additive (new `slice` command + flags + model field + gradient checkpointing), so it's a `z` bump on my side at release — no need to touch the version here.

plz12345 (commented):
Updated, should be good to go, now.

BTW, the LoRA I trained is this one.

https://civitai.com/models/2698739/ltx-23-whisper-and-soft-spoken-audio-lora

dgrauet (approved):
Both requests addressed and then some — the grad-checkpoint equivalence test correctly avoids the `lora_b=0` trivial-pass trap (random-init both adapters + `assert mx.any(g_c != 0)`), so it actually proves it would catch the silent-no-op adapter. Audio-token tests cover trim/pad/exact/last-token-replication/canonical round-trip, and the preprocess behavior change is now documented in the body. All 5 CI checks green. Additive → z bump at release. 🚀

PR #45, opened 2026-06-20
feat: multi-anchor I2V for `--one-stage` and `--distilled`

plz12345:
### Summary

The single-stage and distilled `generate` paths previously routed `--image`
through `_legacy_single_image()`, which raised `SystemExit` on more than one
anchor or any non-trivial `frame_idx`/`strength`. Multi-anchor I2V was therefore
limited to `--two-stage` / `--two-stages-hq`.

Both `TI2VidOneStagePipeline` and the distilled pipeline already accept `images=`
and route through `combined_image_conditionings` — the CLI guard was the only
thing blocking these modes. This PR removes the guard and passes
`images=args.images` on both paths.

Anchor semantics (unchanged, now reachable on all `generate` modes):
- `frame_idx=0` → `VideoConditionByLatentIndex`: hard-replaces the first latent
  frame (strongly preserved).
- `frame_idx>0` → `VideoConditionByKeyframeIndex`: appends a soft keyframe anchor
  at that temporal position.

### Changes

- **`cli.py`** — remove `_legacy_single_image`; `--one-stage` and `--distilled`
  now pass `images=args.images`. Fix `--image` help text: `frame_idx` is a
  **pixel** frame index (0-based), not a latent index.
- **`keyframe_cond.py`** — `VideoConditionByKeyframeIndex.frame_idx` docstring
  corrected to "pixel frame index (0-based)".
- **`CLAUDE.md`** — new "Multi-Anchor I2V" section documenting the repeatable
  `--image PATH FRAME_IDX STRENGTH` form across all `generate` modes.
- **`uv.lock`** (separate commit) — resync to `0.14.12`; the 0.14.12 release
  bumped the workspace `pyproject` versions but left the lockfile at `0.14.11`.

### Validation

Start+end anchor smoke tests at 512×512×25 (start image @ frame 0, end image @
frame 24):

| Mode | E2E | Frame 0 (hard) | Final frame (soft) | Time |
|------|-----|----------------|--------------------|------|
| `--distilled` | ✅ | reproduces start image | tracks end-anchor pose | 29.5s |
| `--one-stage` | ✅ | reproduces start image | tracks end-anchor pose | 465s |

Soft keyframe anchors are hints, not law — they may drift at longer durations,
but both held cleanly at this length.

### Notes

- `--two-stage` / `--two-stages-hq` behavior is unchanged.
- Purely additive: the only paths affected are ones that previously errored out.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -471,6 +471,36 @@ ltx-2-mlx generate \
 
 Flags: `--two-stage` (Euler), `--two-stages-hq` (res_2s), `--cfg-scale` (default 3.0), `--stg-scale` (default 0.0), `--stage1-steps` (default 30 standard, 15 HQ), `--stage2-steps` (default 3), `--image`.
 
+### Multi-Anchor I2V (`--image` repeatable)
+
+All `generate` modes (`--one-stage`, `--two-stage`, `--two-stages-hq`, `--distilled`) support multiple `--image` flags. Each anchor takes `PATH FRAME_IDX STRENGTH` where `FRAME_IDX` is the **pixel frame index** (0-based; for a 97-frame video the last frame is 96).
+
+- `frame_idx=0` → `VideoConditionByLatentIndex`: hard-replaces the first latent frame (strongly preserved)
+- `frame_idx>0` → `VideoConditionByKeyframeIndex`: appends soft reference tokens at that temporal position
+
+```bash
+# Anchor both ends — model animates the transition
+ltx-2-mlx generate \
+  --prompt "woman stands up from chair and walks to kitchen sink" \
+  --two-stage \
+  --image sitting.jpg 0 1.0 \
+  --image standing.jpg 96 1.0 \
+  --image at_sink.jpg 136 1.0 \
+  -f 137 --frame-rate 24 -o output.mp4
+
+# Seamless loop — same image at start and end
+ltx-2-mlx generate \
+  --prompt "flowing water rippling" \
+  --two-stage \
+  --image frame.jpg 0 1.0 \
+  --image frame.jpg 96 1.0 \
+  -f 97 --frame-rate 24 -o loop.mp4
+```
+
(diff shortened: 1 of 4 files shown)


dgrauet (approved):
LGTM — approving. This removes a port-only divergence rather than introducing one, which is exactly the right direction.

## Verified the central claim

Both target pipelines already supported multi-anchor; the CLI guard was the only blocker, confirmed:

- **one-stage** — `ti2vid_one_stage.py:116,166-172` accepts `images=`, routes through `combined_image_conditionings`, strips appended keyframe tokens at L256.
- **distilled** — `distilled.py:119,177,251` accepts `images=` and conditions in **both** stage 1 and stage 2.

Upstream `--image` is repeatable across all modes and `combined_image_conditionings` is the upstream-iso path, so lifting `_legacy_single_image()` makes the port *more* iso, not less. All five references (def + 2 call sites + 2 invocations) are gone.

## Docstring fix is a genuine correctness fix

`VideoConditionByKeyframeIndex.frame_idx` said "Latent frame index", but `_compute_keyframe_positions` in the same file already documented "Pixel frame index" and the math confirms it: `t_mid = (t_mid + frame_idx) / frame_rate` adds `frame_idx` in pixel space before dividing by frame_rate. Your change makes the two docstrings consistent.

## uv.lock sync is legitimate

The 0.14.12 release did leave the lockfile at 0.14.11 while all four pyprojects moved to 0.14.12 — confirmed. Good catch; unblocks a clean `uv sync` for contributors.

## One non-blocking nit (follow-up, not for this PR)

`ti2vid_one_stage.py:235` — `if self.low_memory and image is None:` frees the VAE encoder. Since the CLI now always passes `image=None` (anchors arrive via `images=`), this branch always fires in low-ram. That's **safe** — anchor encoding finishes at L170, well before this point — and is actually a small memory win for low-ram one-stage I2V. But the `image is None` condition is now misleading; semantically it should be `not resolved_images`. Worth a tidy-up in a later pass, no action needed here.

## Versioning

Purely additive (only previously-erroring paths now work) → `z` bump, 0.14.12 → 0.14.13. Correct classification. Bundling the uv.lock sync is slightly against "one PR = one concern" but it's trivial housekeeping tightly coupled to the release, so I'm fine with it here.

Nice work, and thanks for the smoke tests on both modes.

PR #53, opened 2026-06-28
fix: apply streaming LoRA deltas under the correct block prefix (#52)

### LINKED ISSUE #52: low_ram_streaming silently drops LoRA deltas (BlockLoraSource block_prefix mismatch)
## Summary

In `low_ram_streaming` mode, LoRAs are **silently not applied** — generation output is byte-identical to no-LoRA. The non-streaming path fuses the same LoRA correctly. Root cause is a prefix mismatch in `BlockLoraSource` construction.

## Root cause

`packages/ltx-pipelines-mlx/ltx_pipelines_mlx/_base.py`, in the `low_ram_streaming` branch of `_load_transformer` (the `BlockLoraSource(...)` construction):

```python
BlockLoraSource(
    resolved,
    block_prefix="transformer.transformer_blocks.",   # <-- wrong
    strength=strength,
    sd_ops=LTXV_LORA_COMFY_RENAMING_MAP,
)
```

`BlockLoraSource` keeps a delta only if the **renamed** key starts with `block_prefix`
(`block_streaming.py`: `model_key = sd_ops.apply_to_key(raw_key); if not model_key.startswith(block_prefix): continue`).

But `LTXV_LORA_COMFY_RENAMING_MAP` only strips `diffusion_model.` — it does **not** add `transformer.`:

```
diffusion_model.transformer_blocks.0.attn1.to_k.lora_A.weight
  -> (sd_ops) -> transformer_blocks.0.attn1.to_k.lora_A.weight
```

So no renamed key starts with `transformer.transformer_blocks.` → every delta is skipped → the LoRA contributes nothing.

The non-streaming branch matches on `transformer_blocks.` (via `load_split_safetensors(prefix="transformer.")` + `_fuse_pending_loras`), which is why it works.

## Evidence

Trainer-produced LoRA (`ltx-trainer-mlx`, T2V, rank 32, 1728 tensors), keys remapped through `LTXV_LORA_COMFY_RENAMING_MAP`:

```
keys starting with "transformer.transformer_blocks." : 0/1728
keys starting with "transformer_blocks."             : 1728/1728
```

End-to-end (q4 distilled, same seed), Δ mean abs pixel vs no-LoRA:

```
LoRA + low_ram_streaming : Δ = 0.0   (no "Fusing LoRA" msg; deltas all dropped)
LoRA + normal fusion     : Δ = 51.2  (applied)
```

## Fix

Align `block_prefix` with the `sd_ops` output:

```python
block_prefix="transformer_blocks.",
```

This makes all 1728 deltas match (same prefix the non-streaming fusion already uses to parse `(block_idx, param)`).

## Reproduction

1. Train (or take) a T2V LoRA via `ltx-trainer-mlx`.
2. Generate twice with the same seed: once with `low_ram_streaming=True`, once with `False`.
3. `low_ram_streaming=True` output == no-LoRA output; `False` output is styled.

## Impact

On 32 GB machines, low-RAM streaming is the practical default, so LoRAs appear completely non-functional with no error or warning.

## Workaround

Disable `low_ram_streaming` when applying a LoRA (fits on 32 GB with the q4 base).

Found via LTX Desktop macOS (commit `3f5897a`, v0.14.13).

dgrauet:
Closes #52.

## Problem

In `low_ram_streaming` mode, **every LoRA was silently dropped** — generation output was byte-identical to no-LoRA, while the non-streaming path fused the same LoRA correctly.

The three `BlockLoraSource` call sites passed `block_prefix="transformer.transformer_blocks."`, but `LTXV_LORA_COMFY_RENAMING_MAP` only strips `diffusion_model.` and never adds `transformer.`. So a renamed key is `transformer_blocks.N.<param>.lora_{A,B}.weight` and **no delta ever matched the prefix** → all deltas skipped (`block_streaming.py`: `if not model_key.startswith(block_prefix): continue`).

Direct repro:

```
OLD prefix  has_block(0): False   <- the silent no-op
NEW prefix  has_block(0): True    <- fixed
```

## Fix (single source of truth)

Rather than three magic strings that can drift, introduce `LTXV_LORA_BLOCK_PREFIX = "transformer_blocks."` next to the renaming map — it documents and couples *the namespace the map produces* in one place. All three streaming sites route through it:

- `_base.py` — community `--lora` on `generate` modes (`_pending_loras`)
- `ti2vid_two_stages.py` — distilled-LoRA stage-2 fusion at custom strength
- `ic_lora.py` — IC-LoRA control LoRAs

The `BlockStreamer` model-weight prefix (genuinely `transformer.`-prefixed, from `transformer.safetensors`) is **left untouched**.

## Coverage — all LoRA-bearing pipelines

There are only **3 distinct streaming LoRA paths**; every pipeline inherits one of them:

| Funnel | Site | Inherited by |
|---|---|---|
| `_pending_loras` | `_base.py` | one-stage, two-stage, hq, distilled |
| `_swap_to_distilled_streamer` | `ti2vid_two_stages.py` | two-stages-hq, keyframe, a2v, distilled, one-stage |
| `_fuse_loras` | `ic_lora.py` | ic-lora, hdr-ic-lora, lipdub |

`retake`/`extend` don't expose LoRA. Non-streaming paths were already correct.

## Tests

- New regression test (`test_comfy_renamed_lora_matches_pipeline_prefix`): runs a ComfyUI-named LoRA through the renaming map + prefix constant, asserts `has_block(0)` — False under the old literal, True under the constant. This is the silent-no-op class that bit us in #16/#19, so it gets a contract test rather than just the one-char change.
- Fixed two existing dispatch tests that had **codified the buggy prefix**.
- Corrected the misleading `BlockLoraSource` docstring example.
- 32/32 LoRA + streaming tests pass; ruff clean.

## Validation note

No real `--low-ram --lora` generate was run in this environment (no weights/GPU). A Δ-pixel end-to-end check on one funnel (e.g. `generate --lora … --low-ram`) is worth running before merge — the three funnels now share the same constant, so one suffices.

### DIFF
--- a/packages/ltx-core-mlx/src/ltx_core_mlx/loader/__init__.py
+++ b/packages/ltx-core-mlx/src/ltx_core_mlx/loader/__init__.py
@@ -7,6 +7,7 @@ from ltx_core_mlx.loader.primitives import (
     StateDict,
 )
 from ltx_core_mlx.loader.sd_ops import (
+    LTXV_LORA_BLOCK_PREFIX,
     LTXV_LORA_COMFY_RENAMING_MAP,
     ContentMatching,
     ContentReplacement,
@@ -20,6 +21,7 @@ from ltx_core_mlx.loader.sft_loader import (
 )
 
 __all__ = [
+    "LTXV_LORA_BLOCK_PREFIX",
     "LTXV_LORA_COMFY_RENAMING_MAP",
     "ContentMatching",
     "ContentReplacement",
--- a/packages/ltx-core-mlx/src/ltx_core_mlx/loader/block_streaming.py
+++ b/packages/ltx-core-mlx/src/ltx_core_mlx/loader/block_streaming.py
@@ -51,10 +51,14 @@ class BlockLoraSource:
 
     Args:
         lora_path: Path to LoRA safetensors.
-        block_prefix: Same prefix as the streamer (e.g.
-            ``"transformer.transformer_blocks."``). LoRA keys after
-            optional ``sd_ops`` remapping must start with this prefix
-            and be of the form ``f"{block_prefix}{idx}.{param}.lora_A.weight"``.
+        block_prefix: Prefix the LoRA keys carry **after** optional
+            ``sd_ops`` remapping — keys must start with this and be of the
+            form ``f"{block_prefix}{idx}.{param}.lora_A.weight"``. NOTE this
+            is NOT necessarily the streamer's model-weight prefix:
+            ``LTXV_LORA_COMFY_RENAMING_MAP`` produces ``transformer_blocks.``
+            (no ``transformer.``), so callers using that map must pass
+            ``LTXV_LORA_BLOCK_PREFIX``, not the ``transformer.``-prefixed
+            ``BlockStreamer`` prefix. Mixing them silently drops every delta (#52).
         strength: LoRA fusion strength (default 1.0).
         sd_ops: Optional :class:`SDOps` to remap raw safetensors keys
             (e.g. ComfyUI/diffusers → MLX naming via
--- a/packages/ltx-core-mlx/src/ltx_core_mlx/loader/sd_ops.py
+++ b/packages/ltx-core-mlx/src/ltx_core_mlx/loader/sd_ops.py
@@ -146,6 +146,18 @@ LTXV_LORA_COMFY_RENAMING_MAP = (
     .with_replacement("audio_ff.net.2.", "audio_ff.proj_out.")
 )
 
+# Block-key prefix that ``LTXV_LORA_COMFY_RENAMING_MAP`` produces.
+#
+# The map strips ``diffusion_model.`` but never adds ``transformer.``, so a
+# renamed key is ``transformer_blocks.N.<param>.lora_{A,B}.weight``. The
+# streaming LoRA path (``BlockLoraSource``) keeps a delta only if its renamed
+# key starts with the ``block_prefix`` it is given — so every streaming call
+# site MUST pass this constant, not the ``transformer.``-prefixed model-weight
+# prefix used by ``BlockStreamer`` (whose keys come from ``transformer.safetensors``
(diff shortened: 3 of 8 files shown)


dgrauet:
## ✅ End-to-end validation (the reservation in the PR body is now lifted)

Ran a 4-way Δ-pixel test on **M2 Pro 32 GB**, `dgrauet/ltx-2.3-mlx-q8`, `--distilled`, 256×384×9, seed 12345, Union Control LoRA (`diffusion_model.transformer_blocks.*` ComfyUI format → goes through `LTXV_LORA_COMFY_RENAMING_MAP`). This exercises **funnel 1** (community `--lora` via `_base.py`). In distilled mode A and B both perform the distilled-streamer swap, so the *only* difference between them is the community LoRA — Δ(B,A) isolates the fix cleanly.

| Run | Config | Result |
|---|---|---|
| **A** | `--low-ram`, no LoRA | baseline |
| **B_old** | `--low-ram --lora` on the **buggy prefix** | **byte-identical to A** — the silent no-op |
| **B** | `--low-ram --lora` (**fixed**) | **byte-identical to C** |
| **C** | `--lora`, no streaming (known-correct reference) | reference |

```
Δmean=0.000000  Δmax=0.0000   B_old(buggy) vs A(no-LoRA)   -> 0  == silent no-op (the BUG)
Δmean=0.199850  Δmax=0.9608   B    (fixed) vs A(no-LoRA)   -> large == LoRA now applies
Δmean=0.000000  Δmax=0.0000   B    (fixed) vs C(non-stream) -> 0  == streaming == non-streaming
byte-identical: B_old == A  (True),  B == C  (True)
```

Conclusion: pre-fix the streaming LoRA was provably a no-op (output byte-identical to no-LoRA); post-fix the streaming path is byte-identical to the non-streaming reference and differs substantially from no-LoRA. The buggy-prefix run was a temporary local edit, reverted immediately — the committed tree is the fix only.

PR #60, opened 2026-07-05
fix: support quantized transformers at any group_size (not just 64)

plz12345:
## Summary

The runtime assumed a quantization **group size of 64** in two places, so any quantized transformer using a different group size (e.g. an **int4 / group_size 32** build) failed to load or fuse LoRAs. Both spots now derive the true `(bits, group_size)` from shapes that carry it, so any group size (32 / 64 / 128) and bit width (int4 / int8) work.

## Why the tensors alone weren't enough

For a quantized `Linear`, the saved tensors give:

- `packed_cols = in_features * bits / 32`
- `scales_cols = in_features / group_size`

That only pins down `group_size * bits`, not the split — an int4/g32 layer is indistinguishable from int2/g64 without knowing `in_features`. The fix recovers `in_features` from a shape that *does* carry it:

- **`fuse_loras._fuse_delta_with_quantized`** — the LoRA delta has the true `(out, in)` shape. Previously it assumed g64, dequantized to the wrong `in_features`, and raised `[broadcast_shapes] Shapes (O, 2*I) and (O, I) cannot be broadcast` when adding the delta.
- **`weights.apply_quantization`** — the model's float `Linear` weight still has its real `(out, in)` shape at quantize time. Previously `_detect_quantization_bits` assumed g64 and mis-derived bits (int4/g32 → int2/g64), so the rebuilt `QuantizedLinear` had the wrong packed shape and `load_weights` failed with `Expected shape ... but received shape ...`.

Fixing `apply_quantization` covers all five call sites centrally, including `--low-ram` block streaming (which reuses `apply_loras`).

## Behavior change

None for existing group_size=64 models — the derivation returns exactly `(bits, 64)` for them (covered by tests). It only *adds* support for other group sizes.

## Testing

Self-contained regression tests (no on-disk model required, so they run in CI — unlike the existing `@skip_no_weights` integration tests):

- `tests/test_quantization_params.py` — `_derive_quant_params` recovery + strict `load_weights` roundtrip across **int4/int8 × g32/g64/g128**.
- `tests/test_loader.py::TestQuantizedLoraFusion` — quantized LoRA fusion shape preservation + numeric reconstruction across group sizes.

Verified these tests **fail on the pre-fix code with the exact production errors** (`[broadcast_shapes]` and the `load_weights` shape mismatch) while the g64 cases stay green, then pass once the fixes are applied. Full suite: **545 passed, 22 skipped**.

## Motivation

Surfaced converting a distilled LTX-2.3 transformer to **int4 / group_size 32** (int4 benefits from the smaller group size) for a quality-vs-int8 comparison. Nothing published used a non-64 group size, so the assumption had gone unnoticed and untested.

### DIFF
--- a/packages/ltx-core-mlx/src/ltx_core_mlx/loader/fuse_loras.py
+++ b/packages/ltx-core-mlx/src/ltx_core_mlx/loader/fuse_loras.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 import mlx.core as mx
 
 from ltx_core_mlx.loader.primitives import LoraStateDictWithStrength, StateDict
+from ltx_core_mlx.utils.weights import derive_quant_params
 
 
 def apply_loras(
@@ -185,24 +186,15 @@ def _fuse_delta_with_quantized(
     scales = model_sd.sd[scales_key] if scales_key else None
     biases = model_sd.sd.get(biases_key) if biases_key else None
 
-    # Infer quantization parameters BEFORE dequantizing.
-    # MLX packs N-bit values into 32-bit ints: pack_factor = 32 / bits.
-    # For a weight (O, packed) and scales (O, num_groups):
-    #   real_features = num_groups * group_size
-    #   packed_features = real_features / pack_factor
-    # So: pack_factor = real_features / packed_features
-    #     bits = 32 / pack_factor
+    # Infer quantization parameters BEFORE dequantizing. The LoRA delta carries
+    # the true (out, in) weight shape, so the real in_features disambiguates
+    # (bits, group_size) for any group size (32/64/128) and bit width (4/8) —
+    # see derive_quant_params, the single home shared with load-time quantization.
     in_features_packed = weight.shape[-1]
     if scales is not None and scales.ndim > 1:
         num_groups = scales.shape[1]
-        # The LoRA delta carries the true (out, in) weight shape, so derive the
-        # real in_features from it rather than assuming a group size. This keeps
-        # fusion correct for any group_size (32/64/128) and bit width (4/8).
         in_features_real = deltas.shape[-1]
-        # packed cols hold in_features_real * bits / 32 values → solve for bits.
-        bits = max(2, min(8, round(32 * in_features_packed / in_features_real)))
-        # in_features_real is spread evenly across num_groups groups.
-        group_size = in_features_real // num_groups
+        bits, group_size = derive_quant_params(in_features_real, in_features_packed, num_groups)
     else:
         group_size = 64
         bits = 8
--- a/packages/ltx-core-mlx/src/ltx_core_mlx/loader/fuse_loras.py
+++ b/packages/ltx-core-mlx/src/ltx_core_mlx/loader/fuse_loras.py
@@ -195,11 +195,14 @@ def _fuse_delta_with_quantized(
     in_features_packed = weight.shape[-1]
     if scales is not None and scales.ndim > 1:
         num_groups = scales.shape[1]
-        # Try group_size=64 (default), compute real features, check consistency
-        group_size = 64
-        in_features_real = num_groups * group_size
-        pack_factor = in_features_real / in_features_packed if in_features_packed > 0 else 4
-        bits = max(2, min(8, round(32 / pack_factor)))
+        # The LoRA delta carries the true (out, in) weight shape, so derive the
+        # real in_features from it rather than assuming a group size. This keeps
+        # fusion correct for any group_size (32/64/128) and bit width (4/8).
+        in_features_real = deltas.shape[-1]
+        # packed cols hold in_features_real * bits / 32 values → solve for bits.
+        bits = max(2, min(8, round(32 * in_features_packed / in_features_real)))
+        # in_features_real is spread evenly across num_groups groups.
+        group_size = in_features_real // num_groups
     else:
         group_size = 64
         bits = 8
--- a/packages/ltx-core-mlx/src/ltx_core_mlx/utils/weights.py
+++ b/packages/ltx-core-mlx/src/ltx_core_mlx/utils/weights.py
@@ -67,6 +67,50 @@ def _detect_quantization_bits(
     return 8  # default fallback
 
 
+def derive_quant_params(
+    in_features: int,
+    packed_cols: int,
+    scales_cols: int,
+) -> tuple[int, int]:
+    """Derive ``(bits, group_size)`` from a quantized layer's true in_features.
+
+    Given the layer's real ``in_features`` (from the float weight or the LoRA
+    delta, both of which carry the unpacked shape) and the saved packed weight /
+    scales column counts:
+
+        packed_cols = in_features * bits / 32   -> bits = 32 * packed_cols / in_features
+        scales_cols = in_features / group_size  -> group_size = in_features / scales_cols
+
+    The result is validated for *exact* consistency: a stale or mis-shaped tensor
+    whose ratio merely rounds to a plausible ``bits`` is rejected here instead of
+    silently misinterpreting the weights downstream. Single home for the
+    derivation shared by load-time quantization and LoRA fusion.
+
+    Raises:
+        ValueError: if the three counts are not exactly consistent with a valid
+            MLX (bits, group_size) split.
+    """
+    if in_features <= 0 or packed_cols <= 0 or scales_cols <= 0:
+        raise ValueError(
+            "quant param derivation needs positive shapes, got "
+            f"in_features={in_features}, packed_cols={packed_cols}, scales_cols={scales_cols}"
+        )
+    bits = round(32 * packed_cols / in_features)
+    if not 2 <= bits <= 8 or packed_cols * 32 != in_features * bits:
+        raise ValueError(
+            f"packed weight cols ({packed_cols}) are not an exact bit-packing of "
+            f"in_features={in_features}: 32*{packed_cols} != {in_features}*bits for any "
+            f"bits in [2, 8] (nearest bits={bits})"
+        )
+    if in_features % scales_cols != 0:
+        raise ValueError(
+            f"in_features={in_features} is not divisible by scales cols ({scales_cols}); "
+            "cannot derive a uniform group_size"
+        )
+    group_size = in_features // scales_cols
+    return bits, group_size
+
+
 def _derive_quant_params(
     model: nn.Module,
     weights: dict[str, mx.array],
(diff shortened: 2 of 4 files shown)


dgrauet (commented):
Reviewed the derivation approach in depth (traced all five `apply_quantization` call sites, the `--low-ram` path, and the `ic_lora` re-quantization path). The core fix is sound and the test coverage across int4/int8 × g32/g64/g128 is solid — the derivation path engages in every internal load path, and the g64 fallback branch is unreachable in practice. Five findings inline, roughly ordered by severity: one public-API trap, two robustness hardenings, one dedup suggestion, and a nit.

packages/ltx-core-mlx/src/ltx_core_mlx/utils/weights.py
```diff
@@ -95,8 +129,17 @@ def apply_quantization(
     if not quantized_layers:
         return
 
+    # Derive bits AND group_size from the model's true in_features. The saved
+    # tensors alone are ambiguous — packed weight cols and scales cols only pin
+    # down (group_size * bits), not the split — so a g32/int4 model is
+    # indistinguishable from g64/int2 without external truth. The model's float
+    # weights still carry the real (out, in) shape, which disambiguates.
     if bits is None:
```
dgrauet: Passing an explicit `bits` skips group_size derivation entirely: the whole derived block is gated on `if bits is None`, so `apply_quantization(model, weights, bits=4)` on an int4/g32 checkpoint leaves `group_size=64` and `load_weights` fails with a shape mismatch. No internal caller passes `bits` today, but it's a documented public kwarg — consider deriving `group_size` independently of whether `bits` was supplied (or dropping the kwargs now that both are derived).

packages/ltx-core-mlx/src/ltx_core_mlx/utils/weights.py
```diff
@@ -67,6 +67,40 @@ def _detect_quantization_bits(
...
+    for layer in quantized_layers:
+        wkey = f"{layer}.weight"
+        skey = f"{layer}.scales"
+        if wkey in weights and skey in weights and wkey in model_params:
+            in_features = int(model_params[wkey].shape[-1])
+            packed_cols = int(weights[wkey].shape[-1])
+            scales_cols = int(weights[skey].shape[-1])
+            if in_features == 0 or scales_cols == 0:
+                continue
+            bits = max(2, min(8, round(32 * packed_cols / in_features)))
```
dgrauet: No exact-consistency validation on the derived params: nothing asserts `packed_cols * 32 == in_features * bits`. A stale/mis-shaped input whose `in_features` is e.g. half the true value derives an internally shape-consistent but wrong `(bits, group_size)`, so `mx.dequantize` doesn't raise and the weights are silently misinterpreted. A cheap exact-divisibility assert here (and in `_fuse_delta_with_quantized`) would fail fast with a clear message instead — the `max(2, min(8, ...))` clamp currently coerces inconsistent ratios into plausible-looking values.

packages/ltx-core-mlx/src/ltx_core_mlx/utils/weights.py
```diff
@@ -67,6 +67,40 @@ def _detect_quantization_bits(
...
+    from mlx.utils import tree_flatten
+
+    model_params = dict(tree_flatten(model.parameters()))
+    for layer in quantized_layers:
+        wkey = f"{layer}.weight"
+        skey = f"{layer}.scales"
+        if wkey in weights and skey in weights and wkey in model_params:
+            in_features = int(model_params[wkey].shape[-1])
+            packed_cols = int(weights[wkey].shape[-1])
+            scales_cols = int(weights[skey].shape[-1])
```
dgrauet: Latent trap: if the matched model layer is already an `nn.QuantizedLinear` (e.g. the `ic_lora.py` re-quantization path, where the dit is quantized before `_fuse_loras` runs), `model_params[wkey].shape[-1]` is the *packed* column count, not the real in_features — this derives garbage like `(bits=8, group_size=16)`. Today the result is discarded because the `isinstance(_module, nn.Linear)` predicate skips already-quantized layers, so there's no live bug, but a guard here (skip layers whose module is already `QuantizedLinear`) would keep a future refactor from activating it.

packages/ltx-core-mlx/src/ltx_core_mlx/loader/fuse_loras.py
```diff
@@ -195,11 +195,14 @@ def _fuse_delta_with_quantized(
...
-        group_size = 64
-        in_features_real = num_groups * group_size
-        pack_factor = in_features_real / in_features_packed if in_features_packed > 0 else 4
-        bits = max(2, min(8, round(32 / pack_factor)))
+        # The LoRA delta carries the true (out, in) weight shape, so derive the
+        # real in_features from it rather than assuming a group size. This keeps
+        # fusion correct for any group_size (32/64/128) and bit width (4/8).
+        in_features_real = deltas.shape[-1]
+        # packed cols hold in_features_real * bits / 32 values → solve for bits.
+        bits = max(2, min(8, round(32 * in_features_packed / in_features_real)))
```
dgrauet: This derivation math (`bits = max(2, min(8, round(32 * packed / in_features)))`, `group_size = in_features // num_groups`) is duplicated verbatim from `_derive_quant_params` in `utils/weights.py`. Same formula, same three inputs — a shared `derive_quant_params(in_features, packed_cols, scales_cols) -> (bits, group_size)` helper in `utils/weights.py` is feasible with no circular import, and would give the consistency validation a single home so LoRA fusion and load-time quantization can't drift apart.

tests/test_loader.py
```diff
@@ -220,6 +221,81 @@ def test_skips_scales_biases_keys(self):
...
+class TestQuantizedLoraFusion:
+    """Fusing a LoRA delta into a *quantized* weight (dequant -> add -> requant).
+
+    Regression coverage for the group_size!=64 bug in
+    _fuse_delta_with_quantized: it assumed group_size=64, so fusing into an
+    int4/g32 weight dequantized to the wrong in_features and raised
+    "[broadcast_shapes] Shapes (O,2*I) and (O,I) cannot be broadcast".
+    """
+
+    def _quantized_model_sd(self, weight, bits, group_size):
```
dgrauet: Nit: the new helpers/tests in this class lack type hints (repo CLAUDE.md: "Mandatory type hints on all functions"). The pre-existing tests in this file are also unannotated so this matches the neighborhood, but the PR's other new file `test_quantization_params.py` is fully annotated — worth aligning.

packages/ltx-core-mlx/src/ltx_core_mlx/utils/weights.py
plz12345: Fixed in 48cd104: `group_size` is now derived independently of whether `bits` was passed. `apply_quantization` always runs `_derive_quant_params`; an explicit `bits` only overrides the derived bit width, while the derived `group_size` is always applied. So `apply_quantization(model, weights, bits=4)` on an int4/g32 checkpoint now picks up group_size=32 instead of leaving the default 64.

packages/ltx-core-mlx/src/ltx_core_mlx/utils/weights.py
plz12345: Fixed in 48cd104: the derivation now lives in a shared `derive_quant_params()` that asserts exact consistency — `packed_cols * 32 == in_features * bits` and `in_features % scales_cols == 0` — and raises `ValueError` instead of clamping. Added `test_rejects_inexact_bit_packing` / `test_rejects_indivisible_group_size` covering the mis-shaped-input case.

packages/ltx-core-mlx/src/ltx_core_mlx/utils/weights.py
plz12345: Fixed in 48cd104: `_derive_quant_params` now skips any layer whose model module is already quantized (detected via `f'{layer}.scales' in model_params`), so the packed-cols-as-in_features trap can't be activated by a future refactor. Covered by `test_skips_already_quantized_model_layer`.

packages/ltx-core-mlx/src/ltx_core_mlx/loader/fuse_loras.py
plz12345: Fixed in 48cd104: the duplicated math is gone — `_fuse_delta_with_quantized` now calls the shared `derive_quant_params(in_features, packed_cols, scales_cols)` in `utils/weights.py` (no circular import). LoRA fusion and load-time quantization now share one derivation + one consistency check.

tests/test_loader.py
plz12345: Fixed in 48cd104: added type hints to the new `TestQuantizedLoraFusion` helpers and tests (`_quantized_model_sd`, `_lora`, `test_shapes_preserved`, `test_reconstructs_delta_g32`).

dgrauet:
## Review recap ✅

Reviewed the full diff and ran the unit suites (`test_quantization_params.py` + `test_loader.py`): **52/52 pass**.

**Verdict: clean, well-reasoned fix. Mergeable.**

### The fix is correct
The root cause is stated exactly right: the saved tensors only pin down `group_size * bits` (`packed_cols = in*bits/32`, `scales_cols = in/group_size`), not the split — so an int4/g32 layer is indistinguishable from int2/g64 without the true `in_features`. Recovering it from a shape that *does* carry it (the LoRA delta's `(out, in)` in fusion; the model's float weight in `apply_quantization`) is the right disambiguator.

`derive_quant_params` is nicely hardened:
- `bits` is validated for **exact** bit-packing (`packed_cols*32 == in_features*bits`), not just `round()` + clamp — a stale/mis-shaped tensor that merely rounds to a plausible `bits` is rejected instead of silently misread. 👍
- `group_size` divisibility is checked explicitly.

Two subtle points I verified:
- `_derive_quant_params` correctly **skips already-quantized model layers** (`skey in model_params`), where `.weight`'s last dim is the packed column count, not `in_features` — this covers the ic-lora re-quantization path where the DiT is quantized before fusion runs.
- In `apply_quantization`, `group_size` is derived independently of whether an explicit `bits` was passed, which is exactly what prevents `bits=4` on a g32 checkpoint from leaving the default `group_size=64` and failing `load_weights`.

### Test coverage
Parametrized over `(4,32) / (8,64) / (4,64) / (8,32) / (4,128)` — covers the reported int4/g32 failure plus g128 and controls: strict `load_weights` round-trip, delta reconstruction (int8/g32), the rejection cases, the already-quantized skip, and the no-scales no-op. Comprehensive.

### Note on e2e
No end-to-end run here: the repo's own q4 and q8 builds are both `group_size=64` (unaffected by the bug either way), so there's no g32 checkpoint locally to reproduce it against. The fix targets externally-produced non-g64 builds, which is exactly why the synthetic unit tests (quantize tiny models at each g/bits combo) are the right validation surface — and they pass.

Suggested merge order with the other open PRs: **#60 → #63 → #61** (this is the foundational one; little file overlap between #61 and #63).

PR #61, opened 2026-07-06
feat: Prompt Relay temporal prompt gating

plz12345:
## What

Ports [WhatDreamsCost](https://github.com/WhatDreamsCost/WhatDreamsCost-ComfyUI) / Kijai **Prompt Relay** to ltx-2-mlx: sequence local prompts over time within a single generation. A training-free additive Gaussian penalty on the **video→text cross-attention (`attn2`)** gates each local prompt's token range to a slice of the timeline. The global `--prompt` still applies to every frame; each `--segment` biases its own window.

Method: <https://gordonchen19.github.io/Prompt-Relay/>

## Usage

```bash
ltx-2-mlx generate --distilled --frame-rate 24 \
  --prompt "cinematic film, a red-haired woman in a bedroom" \
  --segment "sitting on the bed" \
  --segment "standing and turning to the window" \
  --segment "looking back over her shoulder" \
  -o out.mp4
```

- `--segment "text" [LEN]` — repeatable, timeline order. `LEN` is optional (latent frames); omit to auto-distribute. Pinned + auto beats mix: pinned beats keep their length, the remainder is spread across the auto beats.
- `--relay-epsilon` (falloff, default `1e-3`), `--relay-strength` (penalty multiplier, default `1.0`).

## How it works

- New `conditioning/prompt_relay.py`: token-range mapping (incremental tokenization against the front-packed Gemma connector output), segment-length distribution, and the `(1,1,Nv,Nk)` additive bias.
- Threads a new `video_cross_attention_mask` kwarg through `LTXModel → BasicAVTransformerBlock → attn2` (previously unmasked) and the three denoise loops.
- `BasePipeline._prompt_relay_setup` / `_prompt_relay_mask` are shared by all generate modes (distilled, one-stage, two-stage, two-stage-hq); the mask is rebuilt per stage because tokens-per-frame differs between half/full res.
- **CFG-safe**: on `guided_denoise_loop` / `res2s_denoise_loop` the mask is applied to the positive-context passes (cond/ptb/mod) only and never to the unconditional pass — matching the reference, which never masks the negative prompt.
- Guarded against modality tiling (the mask can't be tiled across a split `Nv`).

## Testing

`tests/test_prompt_relay.py` (8 tests): mask math, token-range mapping, mixed pinned/auto lengths, a block-level test proving the mask reaches `attn2` and changes its output, and a guided-loop test proving cond-only masking under CFG.

- **Distilled** path verified end-to-end (real generation, both stages).
- **Dev / two-stage / HQ** paths are unit- and inspection-verified (cond-only routing test + signature/import checks); I don't have the dev model locally to run them end-to-end, so an extra pair of eyes on those wirings is welcome.

## Scope / limitations

- Video only for now — audio `attn2` is left unmasked (clean follow-up).
- Not compatible with `--tile-frames` / `--tile-spatial`.

### DIFF
--- a/packages/ltx-core-mlx/src/ltx_core_mlx/conditioning/prompt_relay.py
+++ b/packages/ltx-core-mlx/src/ltx_core_mlx/conditioning/prompt_relay.py
@@ -64,6 +64,7 @@ def map_token_ranges(
     tokenizer,
     global_prompt: str,
     local_prompts: list[str],
+    max_length: int | None = None,
 ) -> tuple[str, list[tuple[int, int]]]:
     """Build the combined prompt and each local prompt's ``[start, end)`` token range.
 
@@ -75,6 +76,10 @@ def map_token_ranges(
         tokenizer: mlx-lm tokenizer (``pipe.prompt_encoder._text_encoder._tokenizer``).
         global_prompt: Prompt applied to every frame.
         local_prompts: Ordered per-segment prompts, gated to their time windows.
+        max_length: Encoder sequence limit (``LTX2_GEMMA_MAX_LENGTH``). When set,
+            a combined prompt exceeding it is rejected: the encoder left-truncates
+            (keeps the last ``max_length`` tokens), which shifts every column and
+            makes the ranges point at the wrong tokens — unrecoverable, so fail fast.
 
     Returns:
         ``(combined_prompt, token_ranges)`` where ``combined_prompt`` is what the
@@ -106,6 +111,13 @@ def map_token_ranges(
         token_ranges.append((prev_len, cur_len))
         prev_len = cur_len
 
+    if max_length is not None and len(tokenizer.encode(combined_prompt)) > max_length:
+        raise ValueError(
+            f"Prompt Relay combined prompt exceeds the encoder max length "
+            f"({max_length}); the encoder left-truncates, which would misalign "
+            "every segment's token range. Shorten the prompts/segments or raise "
+            "LTX2_GEMMA_MAX_LENGTH."
+        )
     return combined_prompt, token_ranges
 
 
@@ -159,7 +171,9 @@ def _build_segments(
         if length <= 0:
             frame_cursor += length
             continue
-        midpoint = (2 * frame_cursor + length) / 2.0
+        # Floor-divide to match the reference (WhatDreamsCost build_segments):
+        # an odd-length segment keeps one fully-free anchor frame at its center.
+        midpoint = float((2 * frame_cursor + length) // 2)
         window = max(length // 2 - 2, 0)
         segments.append(
             _Segment(
@@ -207,7 +221,27 @@ def build_relay_mask(
         where a text range is penalised for a video frame outside its window.
     """
     # Reference uses a constant sigma = 1/ln(1/epsilon) regardless of segment length.
-    sigma = 1.0 / math.log(1.0 / epsilon) if 0.0 < epsilon < 1.0 else 0.1448
+    # Reject out-of-range epsilon rather than silently substituting a default: e.g.
+    # epsilon=1.0 means "no falloff" (sigma -> inf) but a fallback would produce the
+    # opposite (sharp default gating).
+    if not 0.0 < epsilon < 1.0:
+        raise ValueError(f"Prompt Relay epsilon must be in (0, 1), got {epsilon}")
+    sigma = 1.0 / math.log(1.0 / epsilon)
+    # A zero-length beat has no temporal window; leaving its columns at cost 0 would
+    # let it attend to *every* frame — the inverse of gating. Fail fast instead.
+    if any(length <= 0 for length in segment_lengths):
+        raise ValueError(
+            f"Prompt Relay segment lengths {segment_lengths} contain a zero-length "
+            "beat: that segment would attend to every frame instead of being gated. "
+            "Give it an explicit length or drop a segment."
+        )
+    # numpy silently clips out-of-bounds column slices, which would ungate a segment
+    # without warning. Reject ranges past the text axis.
+    if any(end > num_text_tokens for _, end in token_ranges):
+        raise ValueError(
+            f"Prompt Relay token ranges {token_ranges} exceed the text axis "
+            f"(num_text_tokens={num_text_tokens})"
+        )
     segments = _build_segments(token_ranges, segment_lengths, strength)
     cost = np.zeros((num_video_tokens, num_text_tokens), dtype=np.float32)
 
--- a/packages/ltx-core-mlx/src/ltx_core_mlx/conditioning/prompt_relay.py
+++ b/packages/ltx-core-mlx/src/ltx_core_mlx/conditioning/prompt_relay.py
@@ -1,0 +1,236 @@
+"""Prompt Relay: temporal cross-attention biasing for LTX-2.
+
+A training-free technique (https://gordonchen19.github.io/Prompt-Relay/) that lets
+a single generation follow a *sequence* of local prompts over time. A global prompt
+is concatenated with N space-separated local prompts; each local prompt's text-token
+range is softly gated — via an additive Gaussian penalty on the video↔text
+cross-attention (``attn2``) — so it only influences the video frames inside its own
+temporal window. Global-prompt tokens (and the connector's register/padding tokens)
+receive zero penalty and attend freely across all frames.
+
+Ported from Kijai/WhatDreamsCost's ComfyUI ``prompt_relay.py``. This module only
+covers the **video** integer-frame path (audio's scaled non-integer-frame path is a
+follow-on). The output is an additive bias of shape ``(1, 1, Nv, Nk)`` suitable for
+``mx.fast.scaled_dot_product_attention(..., mask=...)``.
+
+Token→position note: the Gemma feature connector front-packs valid tokens to the
+front of the ``Nk``-length sequence (see ``text_encoders/gemma`` embeddings
+connector), so text token *i* (in encode order) lives at sequence position *i*.
+Ranges here are therefore computed by incremental tokenization of the combined
+prompt and used directly as column indices into the ``Nk`` axis.
+"""
+
+from __future__ import annotations
+
+import math
+from dataclasses import dataclass
+
+import mlx.core as mx
+import numpy as np
+
+
+@dataclass
+class PromptRelayInput:
+    """User-facing Prompt Relay request carried through the pipeline.
+
+    Attributes:
+        local_prompts: Ordered per-segment prompts (concatenated after the global
+            ``--prompt`` to form the combined prompt that is actually encoded).
+        segment_lengths: Optional per-segment length in *latent* frames. ``None``
+            auto-distributes evenly across the timeline. Length, if given, must
+            match ``local_prompts``.
+        epsilon: Plateau→penalty falloff (smaller = sharper temporal gating).
+        strength: Global multiplier on the penalty magnitude.
+    """
+
+    local_prompts: list[str]
+    segment_lengths: list[int | None] | None = None
+    epsilon: float = 1e-3
+    strength: float = 1.0
+
+
+@dataclass
+class _Segment:
+    """Per-local-prompt metadata for the temporal penalty."""
+
+    tok_start: int
+    tok_end: int
+    midpoint: float  # in latent-frame units
+    window: float  # half-width of the zero-penalty plateau, in latent frames
+    strength: float
+
+
+def map_token_ranges(
+    tokenizer,
+    global_prompt: str,
+    local_prompts: list[str],
+) -> tuple[str, list[tuple[int, int]]]:
+    """Build the combined prompt and each local prompt's ``[start, end)`` token range.
+
+    Uses incremental tokenization (measuring cumulative prefixes) to sidestep
+    SentencePiece context-dependency, mirroring the reference implementation.
+    The tokenizer is an mlx-lm tokenizer whose ``.encode()`` returns a token-id list.
+
+    Args:
+        tokenizer: mlx-lm tokenizer (``pipe.prompt_encoder._text_encoder._tokenizer``).
+        global_prompt: Prompt applied to every frame.
+        local_prompts: Ordered per-segment prompts, gated to their time windows.
+
+    Returns:
+        ``(combined_prompt, token_ranges)`` where ``combined_prompt`` is what the
+        pipeline must actually encode, and ``token_ranges[i]`` is the half-open
+        column range of ``local_prompts[i]`` in the front-packed text axis.
+    """
+    prefixed_locals = [" " + lp for lp in local_prompts]
+    combined_prompt = global_prompt + "".join(prefixed_locals)
+
+    # Detect whether encode() appends an EOS so we can strip its phantom offset
+    # from cumulative counts (else every local range would shift by +1).
+    eos_adj = 0
+    probe = tokenizer.encode("test")
+    eos_id = getattr(tokenizer, "eos_token_id", None)
+    if probe and ((eos_id is not None and probe[-1] == eos_id) or probe[-1] == 1):
+        eos_adj = 1
+
+    def enc_len(text: str) -> int:
+        return len(tokenizer.encode(text.strip())) - eos_adj
+
+    prev_len = enc_len(global_prompt)
+    token_ranges: list[tuple[int, int]] = []
+    built = global_prompt
+    for plp in prefixed_locals:
+        built += plp
+        cur_len = enc_len(built)
+        if cur_len <= prev_len:
+            raise ValueError(f"Local prompt produced no tokens: '{plp.strip()}'")
+        token_ranges.append((prev_len, cur_len))
+        prev_len = cur_len
+
+    return combined_prompt, token_ranges
+
+
+def distribute_segment_lengths(
+    num_segments: int,
+    latent_frames: int,
+    specified_lengths: list[int | None] | None = None,
+) -> list[int]:
+    """Validate or auto-distribute per-segment lengths (in latent frames).
+
+    ``specified_lengths`` may be ``None`` (all beats auto → even ceil split) or a list
+    matching ``num_segments`` where each entry is either a pinned length or ``None``
+    (auto). Pinned beats keep their length; the leftover timeline is spread evenly across
+    the ``None`` beats. Lengths are then clamped so the running sum never exceeds
+    ``latent_frames``.
+    """
+    if specified_lengths is None:
+        step = -(-latent_frames // num_segments)  # ceil division
+        lengths = [step] * num_segments
+    else:
+        if len(specified_lengths) != num_segments:
+            raise ValueError(
+                f"segment length count ({len(specified_lengths)}) must match number of local prompts ({num_segments})"
+            )
+        auto_idx = [i for i, length in enumerate(specified_lengths) if length is None]
+        if auto_idx:
+            pinned_total = sum(length for length in specified_lengths if length is not None)
+            remaining = max(latent_frames - pinned_total, 0)
+            per_auto = -(-remaining // len(auto_idx)) if remaining > 0 else 0  # ceil split
+            lengths = [per_auto if length is None else length for length in specified_lengths]
+        else:
+            lengths = [int(length) for length in specified_lengths]
+
+    effective: list[int] = []
+    cursor = 0
+    for length in lengths:
+        end = min(cursor + length, latent_frames)
+        effective.append(max(end - cursor, 0))
+        cursor = end
+    return effective
+
+
+def _build_segments(
+    token_ranges: list[tuple[int, int]],
+    segment_lengths: list[int],
+    strength: float,
+) -> list[_Segment]:
+    segments: list[_Segment] = []
+    frame_cursor = 0
+    for (tok_start, tok_end), length in zip(token_ranges, segment_lengths):
+        if length <= 0:
+            frame_cursor += length
+            continue
+        midpoint = (2 * frame_cursor + length) / 2.0
+        window = max(length // 2 - 2, 0)
+        segments.append(
+            _Segment(
+                tok_start=tok_start,
+                tok_end=tok_end,
+                midpoint=midpoint,
+                window=float(window),
+                strength=strength,
+            )
+        )
+        frame_cursor += length
+    return segments
+
+
+def build_relay_mask(
+    token_ranges: list[tuple[int, int]],
+    segment_lengths: list[int],
+    num_video_tokens: int,
+    tokens_per_frame: int,
+    latent_frames: int,
+    num_text_tokens: int,
+    epsilon: float = 1e-3,
+    strength: float = 1.0,
+    dtype: mx.Dtype = mx.bfloat16,
+) -> mx.array:
+    """Build the additive Prompt Relay cross-attention bias.
+
+    Args:
+        token_ranges: Per-segment ``[start, end)`` column ranges (from
+            :func:`map_token_ranges`).
+        segment_lengths: Per-segment length in latent frames (from
+            :func:`distribute_segment_lengths`).
+        num_video_tokens: Actual video token count ``Nv`` at attention time. May
+            exceed ``latent_frames * tokens_per_frame`` when keyframe conditioning
+            appends tokens — those trailing rows receive zero penalty.
+        tokens_per_frame: Video tokens per latent frame (``H*W`` of the latent grid).
+        latent_frames: Number of latent frames ``F``.
+        num_text_tokens: Text sequence length ``Nk`` (e.g. ``LTX2_GEMMA_MAX_LENGTH``).
+        epsilon: Controls plateau→penalty falloff (smaller = sharper gating).
+        strength: Global multiplier on the penalty magnitude.
+        dtype: Output dtype (model runs bf16).
+
+    Returns:
+        Additive bias ``(1, 1, Nv, Nk)`` — ``0`` where attention is free, negative
+        where a text range is penalised for a video frame outside its window.
+    """
+    # Reference uses a constant sigma = 1/ln(1/epsilon) regardless of segment length.
+    sigma = 1.0 / math.log(1.0 / epsilon) if 0.0 < epsilon < 1.0 else 0.1448
+    segments = _build_segments(token_ranges, segment_lengths, strength)
+    cost = np.zeros((num_video_tokens, num_text_tokens), dtype=np.float32)
+
+    rows = np.arange(num_video_tokens)
+    query_frames = (rows // tokens_per_frame).astype(np.float32)
+    # Only the first F*tokens_per_frame rows map to real frames; appended keyframe
+    # tokens (if any) sit beyond that and must attend freely.
+    real = rows < latent_frames * tokens_per_frame
+
+    inv_two_sigma_sq = 1.0 / (2.0 * sigma * sigma)
+    for seg in segments:
+        d = np.abs(query_frames - seg.midpoint)
+        penalty = seg.strength * np.square(np.maximum(d - seg.window, 0.0)) * inv_two_sigma_sq
+        penalty = np.where(real, penalty, 0.0).astype(np.float32)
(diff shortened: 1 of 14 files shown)


packages/ltx-core-mlx/src/ltx_core_mlx/conditioning/prompt_relay.py
```diff
@@ -0,0 +1,236 @@
...
+    segment_lengths: list[int],
+    strength: float,
+) -> list[_Segment]:
+    segments: list[_Segment] = []
+    frame_cursor = 0
+    for (tok_start, tok_end), length in zip(token_ranges, segment_lengths):
+        if length <= 0:
+            frame_cursor += length
+            continue
+        midpoint = (2 * frame_cursor + length) / 2.0
```
dgrauet: **Confirmed port deviation (one character).** The reference computes the midpoint with integer floor division — `midpoint = (2 * frame_cursor + L) // 2` in WhatDreamsCost's `build_segments` — but this line uses `/ 2.0`. Identical for even lengths, +0.5 frame off for odd ones: a length-3 segment loses its fully-free anchor frame. Measured on this branch (default epsilon): per-frame bias `[-53.7, -5.96, -5.96]` — no frame of the segment is penalty-free, where the reference guarantees one (`[-23.9, 0.0, -23.9]`).

Test: `TestOddSegmentMidpointParity::test_odd_length_segment_keeps_a_zero_penalty_anchor_frame` (the even-length control passes on both versions).

```suggestion
        midpoint = float((2 * frame_cursor + length) // 2)
```

packages/ltx-core-mlx/src/ltx_core_mlx/conditioning/prompt_relay.py
```diff
@@ -0,0 +1,236 @@
...
+
+def _build_segments(
+    token_ranges: list[tuple[int, int]],
+    segment_lengths: list[int],
+    strength: float,
+) -> list[_Segment]:
+    segments: list[_Segment] = []
+    frame_cursor = 0
+    for (tok_start, tok_end), length in zip(token_ranges, segment_lengths):
+        if length <= 0:
```
dgrauet, line 171: **Confirmed: a zero-length segment silently attends to ALL frames** — the inverse of gating. Its prompt is still in the combined encoded prompt, but skipping it here leaves its token columns at cost 0.

Reachable from ordinary CLI input: `--segment "a" 5 --segment "b"` on a 5-frame clip → `distribute_segment_lengths` yields `[5, 0]` (your own test asserts this collapse), and 3 auto segments over 4 frames → `[2, 2, 0]`. The args.py `LEN_FRAMES must be positive` check only covers user-typed lengths, never derived ones. The user's second beat becomes a *global* prompt with no warning.

Suggest raising in `build_relay_mask` when any length ≤ 0 (patch in the review body does this). Tests: `TestZeroLengthSegmentUngated` (2 tests).

packages/ltx-core-mlx/src/ltx_core_mlx/conditioning/prompt_relay.py
```diff
@@ -0,0 +1,236 @@
...
+    # Only the first F*tokens_per_frame rows map to real frames; appended keyframe
+    # tokens (if any) sit beyond that and must attend freely.
+    real = rows < latent_frames * tokens_per_frame
+
+    inv_two_sigma_sq = 1.0 / (2.0 * sigma * sigma)
+    for seg in segments:
+        d = np.abs(query_frames - seg.midpoint)
+        penalty = seg.strength * np.square(np.maximum(d - seg.window, 0.0)) * inv_two_sigma_sq
+        penalty = np.where(real, penalty, 0.0).astype(np.float32)
+        cost[:, seg.tok_start : seg.tok_end] = penalty[:, None]
```
dgrauet, line 258: **Confirmed: no bounds check, and the whole range mapping is truncation-blind.** Two stacked issues:

1. `numpy` silently clips this slice: a segment whose range lies past `num_text_tokens` writes nothing and is silently ungated. Verified: `token_ranges=[(20, 30)]` with `Nk=16` produces an all-zero mask, no diagnostic.
2. Upstream of this, `map_token_ranges` computes absolute columns from the **untruncated** combined prompt, but `GemmaLanguageModel.tokenize` keeps the *last* `max_length` tokens (left-truncation, `base_encoder.py:67-68`). Past 1024 tokens every surviving column shifts and the penalty lands on the wrong tokens. Realistic trigger: `--enhance-prompt` (up to 512 generated tokens) + 4-5 segments.

Since misalignment is unrecoverable, fail fast: bounds-assert here, and give `map_token_ranges` a `max_length` param that raises on overflow (wired from `_prompt_relay_setup` with `LTX2_GEMMA_MAX_LENGTH`). Tests: `TestTokenRangeOverflow` (2 tests).

packages/ltx-core-mlx/src/ltx_core_mlx/conditioning/prompt_relay.py
```diff
@@ -0,0 +1,236 @@
...
+        epsilon: Controls plateau→penalty falloff (smaller = sharper gating).
+        strength: Global multiplier on the penalty magnitude.
+        dtype: Output dtype (model runs bf16).
+
+    Returns:
+        Additive bias ``(1, 1, Nv, Nk)`` — ``0`` where attention is free, negative
+        where a text range is penalised for a video frame outside its window.
+    """
+    # Reference uses a constant sigma = 1/ln(1/epsilon) regardless of segment length.
+    sigma = 1.0 / math.log(1.0 / epsilon) if 0.0 < epsilon < 1.0 else 0.1448
```
dgrauet: **Confirmed: out-of-range `--relay-epsilon` is silently discarded.** Any value outside (0,1) — including `1.0`, `0`, negatives — takes the `0.1448` fallback, which is just sigma(1e-3) rounded (`1/ln(1000) = 0.144765…`), i.e. near-default sharp gating. Worst case: `--relay-epsilon 1.0` semantically means *no falloff* (sigma → ∞, penalty → 0) but silently produces the strongest default gating instead — the opposite of the request. No CLI-side validation either.

The guard is fine for dodging `ln(1)=0`, but it should reject with a message rather than substitute. Tests: `TestEpsilonNotSilentlyDiscarded` (2 tests).

packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/ti2vid_two_stages.py
```diff
@@ -606,12 +628,16 @@ def generate_and_save(
...
+        prompt_relay=None,
     ) -> str:
         """Generate two-stage video+audio and save to file.
 
         ``stage1_steps`` defaults to ``None`` so subclasses can apply
         their own default (Euler: 30, HQ res_2s: 15) without being
         overridden by this parent method's signature.
+
+        ``prompt_relay`` (a ``PromptRelayInput``) is forwarded only when set;
+        it is currently honoured by the distilled ``generate_two_stage``.
```
dgrauet: Stale docstring: this claims `prompt_relay` “is currently honoured by the distilled `generate_two_stage`”, but this class's own `generate_two_stage` wires the mask into both of its denoise loops. A reader trusting this might assume relay is a no-op on the Euler two-stage path.

packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/distilled.py
```diff
@@ -137,15 +138,33 @@ def generate_two_stage(  # type: ignore[override]
...
             _materialize(video_embeds, audio_embeds)
         if self.low_memory:
             self.prompt_encoder.free()
             aggressive_cleanup()
 
+        # Per-stage Prompt Relay mask builder. Ranges were computed pre-encode;
+        # the mask is rebuilt each stage because tokens-per-frame (H*W) differs.
+        num_text_tokens = video_embeds.shape[1]
+
+        def _relay_mask(latent_f: int, latent_h: int, latent_w: int, num_video_tokens: int):
```
dgrauet: This `_relay_mask` closure is the right pattern — consider making it the *only* pattern. The other pipelines inline the 7-positional-arg `self._prompt_relay_mask(prompt_relay, relay_token_ranges, F, H, W, latent.shape[1], num_text_tokens)` call at 5 more sites (two_stages ×2, two_stages_hq ×2, one_stage ×1), differing only in half/full-res dims. Having `_prompt_relay_setup` return this closure would collapse every site to the 4 stage-varying args and remove the risk of a silently mis-ordered positional arg producing a wrong-shaped mask in one stage only.

Related nits: the new `prompt_relay` params and `_base.py`'s `_prompt_relay_setup`/`_prompt_relay_mask` are missing type hints (repo CLAUDE.md: “Mandatory type hints on all functions” — `prompt_relay.py` itself is fully annotated, so this is just the pipeline layer).

packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/utils/args.py
```diff
@@ -90,4 +90,52 @@ def __call__(  # type: ignore[override]
...
+                parser.error(f"{option_string}: could not parse LEN_FRAMES from {values[1]!r}: {e}")
+            if length <= 0:
+                parser.error(f"{option_string}: LEN_FRAMES must be positive, got {length}")
+
+        existing = getattr(namespace, self.dest, None) or []
+        existing.append(SegmentInput(text=text, length=length))
+        setattr(namespace, self.dest, existing)
+
+
 __all__ = ["DEFAULT_IMAGE_CRF", "ImageAction", "ImageConditioningInput"]
```
dgrauet: `__all__` wasn't updated with `SegmentAction`/`SegmentInput`. Also, `SegmentAction.__call__` duplicates `ImageAction`'s append-to-dest boilerplate (`existing = getattr(...) or []; existing.append(...); setattr(...)`) — a tiny shared helper would keep the two repeatable-list actions from drifting.

dgrauet (commented):
Deep review of the gating math, with every finding challenged empirically: I wrote 8 adversarial regression tests (pure-CPU, fake tokenizer, no model — same style as `tests/test_prompt_relay.py`) encoding the *expected* behavior. On this branch **7 of 8 fail**; with a minimal 53-line fix (included below) **all 16 pass** — my 8 plus your existing 8, unchanged.

Also verified clean during review (no action needed): the `mask=None` path is bit-exact with pre-PR, all new kwargs are keyword-only so no caller breaks, the uncond pass is correctly never masked in both CFG loops, `--low-ram` works (unified block loop), the EOS probe is correct for the real Gemma tokenizer, `window = length // 2 - 2` is verbatim reference parity, and all conditioning modes append tokens at the tail so the frame-major row mapping holds.

Four confirmed bugs inline (each with its failing test named), plus a few cleanups.

<details>
<summary><b>tests/test_prompt_relay_findings.py</b> — drop into <code>tests/</code>; red on this branch, green with the fixes</summary>

```python
"""Adversarial regression tests for the PR #61 review findings.

Each test encodes the *expected* (correct) behavior. On the PR head a failing
test materializes the finding; a passing test refutes it. Pure-CPU, no model.

Run:  PYTHONPATH=packages/ltx-core-mlx/src pytest test_prompt_relay_findings.py -v
"""

from __future__ import annotations

import numpy as np
import pytest

from ltx_core_mlx.conditioning.prompt_relay import (
    build_relay_mask,
    distribute_segment_lengths,
    map_token_ranges,
)


class _FakeTokenizer:
    """Whitespace tokenizer: prepends a BOS id, one id per word. No EOS.

    Mirrors tests/test_prompt_relay.py's fixture (and Gemma's convention).
    """

    eos_token_id = 99

    def encode(self, text: str) -> list[int]:
        words = text.split()
        return [1] + [1000 + i for i, _ in enumerate(words)]


def _mask_np(**kw) -> np.ndarray:
    """build_relay_mask -> squeezed float32 numpy (Nv, Nk)."""
    defaults = dict(
        num_video_tokens=kw.pop("nv"),
        tokens_per_frame=kw.pop("tpf"),
        latent_frames=kw.pop("frames"),
        num_text_tokens=kw.pop("nk"),
    )
    import mlx.core as mx

    m = build_relay_mask(dtype=mx.float32, **defaults, **kw)
    return np.array(m)[0, 0]


# ---------------------------------------------------------------------------
# Finding 1 — zero-length segments silently attend to ALL frames
# ---------------------------------------------------------------------------


class TestZeroLengthSegmentUngated:
    def test_pinned_fills_clip_second_segment_must_not_go_global(self):
        """--segment "a" 5 --segment "b" on a 5-frame clip -> lengths [5, 0].

        Expected: a segment the user asked to gate must never end up influencing
        every frame. Either raise, or penalize its tokens everywhere. On the PR
        head its token columns keep cost=0 -> it attends globally (inverse of
        gating), silently.
        """
        lengths = distribute_segment_lengths(2, 5, [5, None])
        assert lengths == [5, 0]  # author-intended collapse (existing test)

        try:
            mask = _mask_np(
                token_ranges=[(3, 5), (5, 7)],
                segment_lengths=lengths,
                nv=5,
                tpf=1,
                frames=5,
                nk=16,
            )
        except ValueError:
            return  # correct behavior: zero-length segment rejected with a diagnostic
        seg2_cols = mask[:, 5:7]
        # Correct behavior: segment 2 is NOT free on every frame.
        assert not np.all(seg2_cols == 0.0), (
            "0-length segment 'b' has zero penalty on every frame: it attends "
            "globally instead of being gated (or rejected)."
        )

    def test_auto_split_more_segments_than_it_can_fit(self):
        """3 auto segments over 4 frames -> ceil split [2,2,2] -> clamp [2,2,0]."""
        lengths = distribute_segment_lengths(3, 4)
        assert lengths == [2, 2, 0]

        try:
            mask = _mask_np(
                token_ranges=[(3, 5), (5, 7), (7, 9)],
                segment_lengths=lengths,
                nv=4,
                tpf=1,
                frames=4,
                nk=16,
            )
        except ValueError:
            return  # correct behavior: zero-length segment rejected with a diagnostic
        seg3_cols = mask[:, 7:9]
        assert not np.all(seg3_cols == 0.0), (
            "3rd auto segment collapsed to length 0 and silently attends to all "
            "4 frames."
        )


# ---------------------------------------------------------------------------
# Finding 2 — no bounds check: token ranges past num_text_tokens silently clip
# ---------------------------------------------------------------------------


class TestTokenRangeOverflow:
    def test_range_beyond_text_axis_must_not_silently_noop(self):
        """A segment whose range lies past Nk writes nothing (numpy clips).

        Trigger: combined prompt longer than LTX2_GEMMA_MAX_LENGTH (e.g.
        --enhance-prompt + several segments). The encoder left-truncates to the
        last 1024 tokens; map_token_ranges knows nothing about it, so tail
        segments get ranges >= Nk. Expected: an error (misalignment is
        unrecoverable), not a silent no-op that leaves the segment ungated.
        """
        try:
            mask = _mask_np(
                token_ranges=[(20, 30)],  # entirely past Nk=16
                segment_lengths=[4],
                nv=8,
                tpf=1,
                frames=8,
                nk=16,
            )
        except ValueError:
            return  # correct behavior: overflow rejected
        assert not np.all(mask == 0.0), (
            "segment range (20,30) past Nk=16 was silently clipped to "
            "nothing -> segment fully ungated, no diagnostic"
        )

    def test_map_token_ranges_is_truncation_blind(self):
        """map_token_ranges yields ranges past max_length with no error/flag.

        The Gemma encoder keeps the LAST max_length tokens (left-truncation,
        base_encoder.tokenize), shifting every surviving column. Correct
        behavior: map_token_ranges (or its caller) must detect combined-prompt
        overflow. Fake max_length=8 to keep the test small.
        """
        max_length = 8
        tok = _FakeTokenizer()
        global_prompt = " ".join(f"g{i}" for i in range(10))  # 10 words + BOS = 11 > 8
        # Expected API: the mapper takes the encoder's max_length and raises on
        # overflow (left-truncation makes every column misaligned, which is
        # unrecoverable). PR head: no such parameter exists (TypeError) and
        # ranges silently reference untruncated positions.
        with pytest.raises(ValueError):
            map_token_ranges(tok, global_prompt, ["red car"], max_length=max_length)


# ---------------------------------------------------------------------------
# Finding 3 — midpoint / vs // : odd-length segments lose their free frame
# ---------------------------------------------------------------------------


class TestOddSegmentMidpointParity:
    def test_odd_length_segment_keeps_a_zero_penalty_anchor_frame(self):
        """Reference: midpoint = (2*cursor + L) // 2 -> a length-3 segment has
        one fully-free frame (its center). PR head uses / 2.0 -> midpoint lands
        between frames and NO frame of the segment is penalty-free.
        """
        mask = _mask_np(
            token_ranges=[(3, 5)],
            segment_lengths=[3],
            nv=3,
            tpf=1,
            frames=3,
            nk=8,
        )
        seg_col = mask[:, 3]  # per-frame penalty for the segment's first token
        assert np.any(seg_col == 0.0), (
            f"length-3 segment has no zero-penalty frame (per-frame bias: "
            f"{seg_col.tolist()}); reference floor-divides the midpoint so its "
            "center frame is free"
        )

    def test_even_length_segment_unaffected_control(self):
        """Control: even lengths agree between / and // — must keep passing."""
        mask = _mask_np(
            token_ranges=[(3, 5)],
            segment_lengths=[2],
            nv=2,
            tpf=1,
            frames=2,
            nk=8,
        )
        seg_col = mask[:, 3]
        assert seg_col[1] == 0.0  # frame 1 is the (floored) midpoint: free


# ---------------------------------------------------------------------------
# Finding 4 — out-of-range epsilon silently replaced by the default sigma
# ---------------------------------------------------------------------------


class TestEpsilonNotSilentlyDiscarded:
    def _mask_for_eps(self, eps: float) -> np.ndarray:
        return _mask_np(
            token_ranges=[(3, 5)],
            segment_lengths=[2],
            nv=6,
            tpf=1,
            frames=6,
            nk=8,
            epsilon=eps,
        )

    def test_epsilon_one_is_not_the_default_gating(self):
        """epsilon -> 1 means sigma -> inf, i.e. no temporal falloff. On the PR
        head --relay-epsilon 1.0 silently produces the SAME mask as the sharp
        default 1e-3 — the opposite of the request. Expected: either honor the
        limit (mask ~ 0) or reject the value; never silently substitute.
        """
        try:
            m_one = self._mask_for_eps(1.0)
        except ValueError:
            return  # rejecting out-of-range epsilon is acceptable correct behavior
        # epsilon -> 1 limit: sigma -> inf, penalty -> 0 everywhere.
        assert np.allclose(m_one, 0.0, atol=1e-3), (
            f"--relay-epsilon 1.0 (= no falloff) silently produced near-default "
            f"sharp gating instead (max |bias| = {np.abs(m_one).max():.1f}); the "
            "hardcoded sigma=0.1448 fallback discarded the user's value"
        )

    def test_negative_epsilon_rejected_or_distinct(self):
        with pytest.raises(ValueError):
            self._mask_for_eps(-0.5)  # nonsensical value must be rejected, not defaulted

```
</details>

<details>
<summary><b>candidate_fixes.patch</b> — minimal fix validated against both suites (the <code>max_length</code> wiring from <code>_prompt_relay_setup</code> is left to you)</summary>

```diff
diff --git a/packages/ltx-core-mlx/src/ltx_core_mlx/conditioning/prompt_relay.py b/packages/ltx-core-mlx/src/ltx_core_mlx/conditioning/prompt_relay.py
index 56b281b..3fc7bb9 100644
--- a/packages/ltx-core-mlx/src/ltx_core_mlx/conditioning/prompt_relay.py
+++ b/packages/ltx-core-mlx/src/ltx_core_mlx/conditioning/prompt_relay.py
@@ -64,6 +64,7 @@ def map_token_ranges(
     tokenizer,
     global_prompt: str,
     local_prompts: list[str],
+    max_length: int | None = None,
 ) -> tuple[str, list[tuple[int, int]]]:
     """Build the combined prompt and each local prompt's ``[start, end)`` token range.
 
@@ -106,6 +107,11 @@ def map_token_ranges(
         token_ranges.append((prev_len, cur_len))
         prev_len = cur_len
 
+    if max_length is not None and len(tokenizer.encode(combined_prompt)) > max_length:
+        raise ValueError(
+            f"combined prompt exceeds the encoder max length ({max_length}); the "
+            "encoder left-truncates, which would misalign every segment's token range"
+        )
     return combined_prompt, token_ranges
 
 
@@ -159,7 +165,7 @@ def _build_segments(
         if length <= 0:
             frame_cursor += length
             continue
-        midpoint = (2 * frame_cursor + length) / 2.0
+        midpoint = float((2 * frame_cursor + length) // 2)
         window = max(length // 2 - 2, 0)
         segments.append(
             _Segment(
@@ -207,7 +213,18 @@ def build_relay_mask(
         where a text range is penalised for a video frame outside its window.
     """
     # Reference uses a constant sigma = 1/ln(1/epsilon) regardless of segment length.
-    sigma = 1.0 / math.log(1.0 / epsilon) if 0.0 < epsilon < 1.0 else 0.1448
+    if not 0.0 < epsilon < 1.0:
+        raise ValueError(f"relay epsilon must be in (0, 1), got {epsilon}")
+    sigma = 1.0 / math.log(1.0 / epsilon)
+    if any(length <= 0 for length in segment_lengths):
+        raise ValueError(
+            f"segment lengths {segment_lengths} contain a zero-length beat: that "
+            "segment would attend to every frame instead of being gated"
+        )
+    if any(end > num_text_tokens for _, end in token_ranges):
+        raise ValueError(
+            f"token ranges {token_ranges} exceed the text axis ({num_text_tokens})"
+        )
     segments = _build_segments(token_ranges, segment_lengths, strength)
     cost = np.zeros((num_video_tokens, num_text_tokens), dtype=np.float32)
 

```
</details>

packages/ltx-core-mlx/src/ltx_core_mlx/conditioning/prompt_relay.py
plz12345: Fixed in b0b67bc — `midpoint = float((2 * frame_cursor + length) // 2)`. `TestOddSegmentMidpointParity` now passes (length-3 beat keeps its free anchor frame; even-length control unchanged).

packages/ltx-core-mlx/src/ltx_core_mlx/conditioning/prompt_relay.py
plz12345: Fixed in b0b67bc — `build_relay_mask` now raises when any segment length ≤ 0. `distribute_segment_lengths` still clamps to `[5, 0]`/`[2, 2, 0]` (existing test unchanged); the failure now surfaces loudly at mask-build instead of silently going global. `TestZeroLengthSegmentUngated` (2) pass.

packages/ltx-core-mlx/src/ltx_core_mlx/conditioning/prompt_relay.py
plz12345: Fixed in b0b67bc — both halves: (1) `build_relay_mask` bounds-checks `end > num_text_tokens` and raises; (2) `map_token_ranges` gained a `max_length` param that rejects a combined prompt the encoder would left-truncate, wired from `_prompt_relay_setup` with `LTX2_GEMMA_MAX_LENGTH`. `TestTokenRangeOverflow` (2) pass.

packages/ltx-core-mlx/src/ltx_core_mlx/conditioning/prompt_relay.py
plz12345: Fixed in b0b67bc — epsilon outside `(0, 1)` now raises instead of falling back to the `0.1448` default. `--relay-epsilon 1.0` no longer silently produces sharp default gating. `TestEpsilonNotSilentlyDiscarded` (2) pass.

packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/ti2vid_two_stages.py
plz12345: Fixed in b0b67bc — docstring now reads "`generate_two_stage` wires the mask into both of its denoise loops."

packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/distilled.py
plz12345: Done in b0b67bc — made it the only pattern. Added `_prompt_relay_mask_builder(prompt_relay, token_ranges, num_text_tokens)` on `_base` returning a closure over the 4 stage-varying dims. All 6 sites (one_stage, two_stages ×2, two_stages_hq ×2, distilled ×2) now call `relay_mask(F, H, W, num_video_tokens)` — no more 7-positional-arg calls, and distilled’s inline closure is gone. Also added the missing type hints on `_prompt_relay_setup`/`_prompt_relay_mask_builder`.

packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/utils/args.py
plz12345: Fixed in b0b67bc — `SegmentAction`/`SegmentInput` added to `__all__`, and both list actions now share a `_append_to_dest(namespace, dest, item)` helper.

plz12345:
All review findings addressed in b0b67bc.

**Four confirmed bugs** (`prompt_relay.py`):
1. Zero-length segment beat → now rejected (was silently attending to every frame).
2. Token ranges past the text axis → bounds-checked + `map_token_ranges` gained a `max_length` guard against the encoder’s left-truncation misalignment (wired from `LTX2_GEMMA_MAX_LENGTH`).
3. Segment midpoint `/ 2.0` → `// 2` for reference parity (odd-length beats keep their free anchor frame).
4. Out-of-range `--relay-epsilon` → rejected instead of silently defaulting sigma.

**Cleanups:**
- Collapsed the 6 per-stage mask sites (incl. distilled’s inline closure) into a single `_prompt_relay_mask_builder` closure taking only the stage-varying dims — one build path, no mis-orderable positional args.
- Type hints on the `_base` prompt-relay helpers; `SegmentAction`/`SegmentInput` added to `args.__all__`; shared `_append_to_dest` for the repeatable list actions; fixed the stale distilled-only docstring.

**Tests:** added `tests/test_prompt_relay_findings.py` (your 8 adversarial tests + 2 in-range controls, adapted). Full suite: **546 passed, 22 skipped**; ruff clean.

dgrauet:
## Review approfondie + validation e2e ✅

Reviewed the full diff (code-path tracing, not just the description) and ran end-to-end validation on M2 Pro 32 GB, q8 + `--low-ram`, tiny res (256×256×9).

**Verdict: mergeable. The feature is byte-for-byte inert on the default path and functional when activated.**

### Correctness — the front-packing assumption holds
The core mechanism relies on "text token *i* lives at column *i*" in the `Nk` axis. This is the subtle trap that would silently misalign the whole mask, so I traced it: the Gemma encoder **left-pads** (valid tokens at the end), but `_replace_padding_with_registers` extracts the valid tokens and **re-packs them to the front** + registers. So token *i* → column *i* is correct. 👍

Full plumbing verified: `LTXModel` / `StreamingLTXModel` (`--low-ram`) / `X0Model` all forward the kwarg; `attn2` passes it to SDPA; the negative pass is forced to `None`; modality tiling is rejected with a clear `ValueError`; `max_length` truncation is rejected loudly.

### e2e evidence
| Test | Result |
|------|--------|
| `test_prompt_relay` + `test_prompt_relay_findings` | **18/18 ✅** |
| **main vs this branch, no `--segment`, same seed** | **SHA-256 byte-identical** → feature 100% inert on reference usage |
| distilled + `--segment` | e2e OK, output differs (mask active) |
| **one-stage CFG + `--segment`** | e2e OK (`guided_denoise_loop`, decode, mux) → exercises the `mask=None` override on the uncond pass |

### Minor notes (non-blocking)
- Token boundaries at segment joins can be slightly imprecise if SentencePiece merges the last global token with the first local one — inherent to the technique, matches the reference.
- `_prompt_relay_setup` calls `_load_text_encoder()` early; confirmed idempotent (0.0 s on the second call) — no Gemma double-load regression.

Nice work — the divergence notes and the fail-fast guards (zero-length beat, epsilon range, token overflow) are exactly the right instinct.

PR #63, opened 2026-07-09
fix: ic-lora dev-mode fuses distilled lora in the main pass; add --single-stage

plz12345:
## Problem

The `ic-lora` pipeline diverged from the reference ComfyUI-LTXVideo IC-LoRA workflows, giving poor output on the Union Control workflow — weak prompt adherence with the distilled checkpoint, and motion/hand distortion when swapping in the dev checkpoint.

Comparing against every `LTX-2.3_ICLoRA_*` example workflow shows one consistent recipe:

- **dev** checkpoint (`ltx-2.3-22b-dev`), not distilled
- task IC-LoRA @ **1.0** **and** `distilled-lora-384-1.1` @ **0.5** fused into the **same** model for the main pass (the distilled LoRA @0.5 is what lets the dev model sample in 8 steps — it is not a stage-2-only thing)
- stage-1 sigmas = the fixed 8-step `DISTILLED_SIGMAS` (ManualSigmas), **not** a dynamic shift schedule
- two-stage variants keep **both** LoRAs across both stages — they never reload a clean model

The previous dev-mode implementation fused the distilled LoRA only in stage 2, substituted a dynamic dev sigma schedule, and silently fell back to distilled mode on a bad `--dev-transformer` — reproducing the exact degraded config.

## Changes

- **`_effective_lora_paths`** appends the distilled LoRA alongside the task IC-LoRA so both fuse together **before stage 1** (LoRA deltas are additive; order-independent). Fixes the motion/hand distortion.
- **Dev-mode two-stage keeps both LoRAs across stages** (skips the clean-transformer reload). Legacy distilled (non-dev) path unchanged.
- **Revert the dynamic dev sigma schedule** — dev and distilled share `DISTILLED_SIGMAS`.
- **Hard-fail on a missing `--dev-transformer`** instead of silently reverting to distilled mode (which would also drop the distilled LoRA).
- **Restore** the `low_ram_streaming` cleanup branch that had been gutted.
- **Add `--single-stage`**: full-res one-pass generation with the reference conditioning applied throughout (Comfy Union Control topology — no upsampler / stage 2). Tracks the control video noticeably tighter than two-stage, since two-stage generates stage 1 at half res and drops the reference in the stage-2 refine.
- CLI: bump `--distilled-lora` default to `384-1.1`; print dev-mode / distilled LoRA and topology in the run summary.

## Validation

- Confirmed end-to-end on a pose (DWPose) Union Control dance clip: dev + both LoRAs gives strong prompt adherence and clean motion; `--single-stage` tracks the control skeleton tightly at full res.
- New unit tests cover `_effective_lora_paths` (distilled / dev / relative / absolute / missing distilled) and dev-transformer detection.

## Notes

- No collision with the other open PRs (#60 quant-fix, #61 prompt-relay): the only shared file is `cli.py` with #61, in fully disjoint regions (`ic-lora` subparser + `_cmd_ic_lora` vs the generate command). `ic_lora.py` is untouched by any open PR.

### DIFF
--- a/packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/cli.py
+++ b/packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/cli.py
@@ -368,6 +368,40 @@ examples:
         help="IC-LoRA conditioning attention strength 0.0-1.0 (default: 1.0)",
     )
     ic.add_argument("--skip-stage-2", action="store_true", help="Skip stage 2 upsampling (half resolution output)")
+    ic.add_argument(
+        "--single-stage",
+        action="store_true",
+        help=(
+            "Single-stage full-resolution generation with the reference conditioning "
+            "applied throughout (Comfy Union Control topology). No upsampler / Stage 2. "
+            "Tracks the control signal more tightly than two-stage; slower at large res."
+        ),
+    )
+
+    ic.add_argument(
+        "--dev-transformer",
+        default=None,
+        help=(
+            "Dev (non-distilled) transformer filename. Enables dev mode: the "
+            "distilled LoRA is fused alongside the IC-LoRA (Comfy IC-LoRA recipe). "
+            "e.g. transformer.safetensors in a dev model dir"
+        ),
+    )
+
+    ic.add_argument(
+        "--distilled-lora",
+        default="ltx-2.3-22b-distilled-lora-384-1.1.safetensors",
+        help=(
+            "Distilled LoRA fused alongside the IC-LoRA in dev mode "
+            "(default: ltx-2.3-22b-distilled-lora-384-1.1.safetensors)"
+        ),
+    )
+    ic.add_argument(
+        "--distilled-lora-strength",
+        type=float,
+        default=0.5,
+        help="Distilled LoRA strength in dev mode (default: 0.5)",
+    )
 
     # --- lipdub ---
     ld = sub.add_parser(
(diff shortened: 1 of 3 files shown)


dgrauet:
## Review approfondie + validation e2e ✅

Reviewed the full diff and ran end-to-end validation on M2 Pro 32 GB, q8 + `--low-ram`, tiny res (256×256×9), with the cached Union-Control IC-LoRA + a synthetic canny control video.

**Verdict: mergeable. Dev mode is opt-in, so the default distilled path is unchanged by construction; both new paths run e2e.**

### e2e evidence
| Test | Result |
|------|--------|
| `test_iclora_dev_lora` | **8/8 ✅** |
| Default (no `--dev-transformer`) | `dev_mode=False` → `_effective_lora_paths` returns the task LoRA only, `load()` unchanged (covered by unit tests) |
| `--single-stage` | **e2e OK** (8-step distilled, no Stage 2, decode, mux) |
| **`--dev-transformer` (the core fix)** | **e2e OK** — dev transformer + **task-LoRA @1.0 + distilled-LoRA @0.5 fused together in the same pass** (bind-time streaming), 8 steps, valid output |

Two confirmations on the dev-mode fix:
- It produces valid pixels end-to-end. Step time is ~15 s vs ~7 s for the single-LoRA distilled path — consistent with the documented slower bind-time dual-LoRA fusion at strength ≠ 1.0.
- Output SHA differs from the distilled single-stage run → the dev config actually engages a different model, not a silent no-op.

### Design notes 👍
The fail-fast on a missing `--dev-transformer` (no silent fallback to distilled, which would drop the distilled LoRA and reproduce the bad-output config), the additive both-LoRAs-in-one-pass fusion, and keeping both LoRAs across stages in dev mode all mirror the Comfy Union-Control recipe cleanly. `--single-stage` taking precedence over `--skip-stage-2` is clearly documented.

Note for reviewers on 32 GB: dev mode + `--low-ram` uses bind-time dual-LoRA fusion (dequant → W+BA → requant per linear/block/step), which is heavy — expect a Metal-watchdog sensitivity on contended machines. Not a code issue (the reference two-stage path hits the same watchdog on a saturated box); a rested machine ran it first try.

PR #68, opened 2026-07-10
feat: ic-lora --upsample-only + control-aware --refine-steps

plz12345:
## Summary

Adds two cheaper IC-LoRA topologies for Apple Silicon, where native single-stage at usable resolutions is prohibitively slow (~30 min for a 20s 480×928 clip on high-end Macs):

- **`--upsample-only`** — half-res generative pass (control applied throughout, same as Stage 1), 2× latent upsample, decode directly. No refine. Fast rough draft.
- **`--refine-steps N`** — with `--upsample-only`, run an N-step *control-aware* refine after the upsample: re-encode the control at full resolution and re-append it with the IC-LoRA kept fused. Cleans upsampler artifacts while the now-full-res control sharpens adherence.

Legacy two-stage and single-stage paths are untouched.

## Why the refine is not just two-stage Stage 2

Same input latent, same resolution, same sigma schedule (`--refine-steps 3` == `STAGE_2_SIGMAS`). The difference is what steers the refine:

| | latent res | sigma schedule (steps=3) | control reference | model |
|---|---|---|---|---|
| two-stage Stage 2 (non-dev) | full | `STAGE_2_SIGMAS` | dropped | clean distilled (no IC-LoRA) |
| `--refine-steps 3` | full | `STAGE_2_SIGMAS` (identical) | re-appended @ full res | distilled + IC-LoRA fused |

Two-stage Stage 2 is a control-*blind* cleanup (reference dropped, clean model reloaded); this is a control-*aware* one. Both changes are needed together — the re-appended reference tokens are only meaningful because the IC-LoRA stays fused to consume them. Bonus: the control is re-encoded at full res for the refine, sharper than the half-res signal Stage 1 saw.

## Refine depth

`--refine-steps` maps onto the distilled sigma tail — N steps ending at 0.0, higher N starts from a higher sigma (more cleanup, more drift from the draft):

| N | start σ | behavior |
|---|---|---|
| 1 | 0.42 | lightest — de-fuzz, closest to draft |
| 3 | 0.91 | matches standard Stage 2 depth |
| 4 | 0.98 | more aggressive limb/artifact reconstruction |
| 8 | 1.0 | full regeneration at full res |

Capped at `len(DISTILLED_SIGMAS) - 1 = 8`.

## Also fixed

Half-res gen dims are floored to a multiple of 32 so image conditioning encodes on a VAE-compatible grid (raw `height // 2` can leave a remainder, e.g. `928 // 2 = 464`, crashing the encoder's `space_to_depth`). Latent-neutral — the latent grid already floors `// 32` — so it also hardens the default two-stage path for image-conditioned runs at non-64-divisible dims.

## Results (480×928 → 448×896 out, 481 frames, seed 42, union-control ref0.5, M5 Pro 64GB Macbook Pro)

| topology | wall time | quality notes |
|---|---|---|
| single-stage (baseline) | ~30 min | reference quality, control-tight |
| `--upsample-only` | ~4.4 min | fast; blurred edges, scattered fast-motion, missing arm (dance sequence) |
| `--upsample-only --refine-steps 3` | ~14 min | close to single-stage; upper-body scatter and missing arm resolved; residual softness on fast-moving hands is genuine motion blur, not refinement artifacts |

Timing breakdown for `--refine-steps 3` (run 1): Stage 1 8×23.09s = 3:04, refine 3×190.84s = 9:32, decode 80.9s → ~14 min (~2.1× faster than single-stage). The refine step is ~8× a Stage 1 step: full-res gen tokens (4×) plus the appended full-res control tokens with O(n²) attention.

Second validation run (different reference image + conditioning, `--refine-steps 3`): Stage 1 3:01, refine 9:06, decode 79.0s → ~13.5 min (consistent). Movements stayed true to the conditioning video — the control-aware refine generalized. Quality was below run 1, but the cause was an input mismatch (the conditioning image was not framed to the control stick figure, forcing a zoom-out that cost face identity), not a refine-mechanism regression. Framing the reference image to the control subject matters for final quality.

## Scope / safety

- New behavior is gated behind `--upsample-only` (+ `--refine-steps > 0`); the default two-stage path is bit-for-bit unchanged (`control_aware_refine=False` preserves the clean-model reload, the `STAGE_2_SIGMAS` schedule, and no-append behavior).
- CI gates pass locally: `ruff check packages/ tests/`, `ruff format --check packages/ tests/`, AST syntax pass. Existing IC-LoRA tests (14) pass.

### DIFF
--- a/packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/cli.py
+++ b/packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/cli.py
@@ -409,6 +409,18 @@ examples:
         ),
     )
     ic.add_argument(
+        "--refine-steps",
+        type=int,
+        default=None,
+        help=(
+            "With --upsample-only, run an N-step control-aware refine after the "
+            "upsample: re-encodes the control at full resolution and re-applies it "
+            "with the IC-LoRA fused, cleaning upsampler artifacts without the "
+            "adherence drift of two-stage. 3 ~ matches the standard Stage 2 depth; "
+            "higher = more cleanup + more drift from the draft (capped at 8)."
+        ),
+    )
+    ic.add_argument(
         "--single-stage",
         action="store_true",
         help=(
@@ -1020,6 +1032,8 @@ def _cmd_ic_lora(args: argparse.Namespace) -> None:
     if not args.quiet:
         if args.single_stage:
             topology = "single-stage full-res"
+        elif args.upsample_only and args.refine_steps:
+            topology = f"upsample + {args.refine_steps}-step control-aware refine (half-res gen + 2x upsample)"
         elif args.upsample_only:
             topology = "upsample-only (half-res gen + 2x upsample, no refine)"
         else:
@@ -1065,6 +1079,7 @@ def _cmd_ic_lora(args: argparse.Namespace) -> None:
         skip_stage_2=args.skip_stage_2,
         single_stage=args.single_stage,
         upsample_only=args.upsample_only,
+        refine_steps=args.refine_steps,
     )
     _print_result(args.output, t0, args.quiet)
 
--- a/packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/cli.py
+++ b/packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/cli.py
@@ -399,6 +399,16 @@ examples:
     )
     ic.add_argument("--skip-stage-2", action="store_true", help="Skip stage 2 upsampling (half resolution output)")
     ic.add_argument(
+        "--upsample-only",
+        action="store_true",
+        help=(
+            "Half-res generative pass (control applied throughout) + 2x latent "
+            "upsample, then decode directly at full resolution — skips the Stage 2 "
+            "refine that drops control adherence. ~3x faster than --single-stage; "
+            "control fidelity sits between two-stage and native single-stage."
+        ),
+    )
+    ic.add_argument(
         "--single-stage",
         action="store_true",
         help=(
@@ -1008,7 +1018,12 @@ def _cmd_ic_lora(args: argparse.Namespace) -> None:
     video_conditioning = [(path, float(strength)) for path, strength in args.video_conditioning]
 
     if not args.quiet:
-        topology = "single-stage full-res" if args.single_stage else "two-stage"
+        if args.single_stage:
+            topology = "single-stage full-res"
+        elif args.upsample_only:
+            topology = "upsample-only (half-res gen + 2x upsample, no refine)"
+        else:
+            topology = "two-stage"
         print(f"Mode: IC-LoRA ({topology})")
         for path, strength in lora_paths:
             print(f"  LoRA: {path} (strength={strength})")
@@ -1049,6 +1064,7 @@ def _cmd_ic_lora(args: argparse.Namespace) -> None:
         conditioning_attention_strength=args.conditioning_strength,
         skip_stage_2=args.skip_stage_2,
         single_stage=args.single_stage,
+        upsample_only=args.upsample_only,
     )
     _print_result(args.output, t0, args.quiet)
 
--- a/packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/ic_lora.py
+++ b/packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/ic_lora.py
@@ -364,6 +364,7 @@ class ICLoraPipeline(BasePipeline):
         conditioning_attention_mask: mx.array | None = None,
         skip_stage_2: bool = False,
         single_stage: bool = False,
+        upsample_only: bool = False,
     ) -> tuple[mx.array, mx.array]:
         """Generate video with IC-LoRA reference conditioning.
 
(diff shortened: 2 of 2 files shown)


dgrauet:
## Review — deep dive + local e2e validation

Reviewed the diff in depth and ran an e2e matrix locally (M2 Pro 32 GB, q8 `--low-ram`, seed 42, Union Control LoRA, synthetic canny control). Everything checks out.

### Code review findings

- **Default-path neutrality**: `control_aware_refine=False` preserves the clean-model reload, `STAGE_2_SIGMAS`, no reference append, and unchanged unpatchify. Confirmed bit-for-bit below.
- **The un-gated `gen_h/gen_w` floor-to-32 change** is the one part that touches the default path — and it's a genuine crash fix, not a behavior change. For any height/width multiple of 64, `(x//2//32)*32 == x//2` (identical). For the rest, the old code fed non-32-multiple dims into the VAE encoder, which crashes in `space_to_depth` — including the **default `--height 480`** (half = 240): reproduced the crash on `main` with `--image` (see e2e B below). The latent grid is unchanged (`compute_video_latent_shape` re-floors `//32`), and the ref-video resolution snaps identically through the same helper, so the claim "latent-neutral" holds.
- **`refine_steps=3 == STAGE_2_SIGMAS`**: verified — `DISTILLED_SIGMAS[-4:] == [0.909375, 0.725, 0.421875, 0.0] == STAGE_2_SIGMAS`.
- **low-ram / dev-mode composition**: skipping `_reload_clean_transformer()` under `control_aware_refine` correctly keeps the streaming `BlockLoraSource`s attached, so the IC-LoRA stays fused to consume the re-appended reference tokens. Composes cleanly with dev mode.
- **Appended-token strip** before unpatchify mirrors the Stage 1 gen-token extraction. Correct.
- Local test suite: 161/162 pass on the branch; the single failure (`test_generate_tiny_video`, ffmpeg pipe) fails identically on `main` → environmental. `ruff check` + `format --check` clean.

### E2E results

| Run | Config | Result |
|---|---|---|
| A (main vs PR) | default two-stage, 384×640×25 (mult-64 dims) | **SHA-256 identical** — default path bit-for-bit unchanged |
| B (main) | `--image` at 480×704×9 (half=240) | **crash on main**: `ValueError: [reshape] Cannot reshape array of size 675840 into shape (1,1,2,7,2,11,2,1024)` in `space_to_depth` |
| B (PR) | same | **OK** (112 s), valid 704×448 output |
| C (PR) | `--upsample-only` 384×640×25 | OK, 99 s (vs 133 s two-stage) |
| D (PR) | `--upsample-only --refine-steps 3` | OK, 157 s, distinct output from both A and C |

No Metal watchdog trips, no memory anomalies on 32 GB.

### Nits (non-blocking)

- No new unit tests — a cheap test for the sigma-tail slicing and the dims flooring would lock these in.
- `--refine-steps` without `--upsample-only` is silently ignored (documented in help, but a CLI warning would be friendlier); same for `--upsample-only` combined with `--single-stage`/`--skip-stage-2`. A negative `--refine-steps` silently clamps to 1.
- At non-mult-64 target dims the output snaps down (requested 480 high → 448 out). Documented trade-off vs the old crash, but a one-line warning when `height % 64 or width % 64` would help users understand the output size.

Nice work — the "why" comments in the refine path made this easy to review.

plz12345:
w/r/t the crash, I planned on opening an issue to discuss better handling of dimension fails like that, but will note it here since you caught it.  My train of thought was 2 options

1. Detect and fail-fast, with a CLI error noting that is what happened and offering rounded-down dims
2. Silently change to non-failure rounded-down values.

@dgrauet what do you think on either option?

dgrauet:
Thanks for raising it — I'd go with **option 2 (round down, as merged) plus an informational warning**, not fail-fast. Three reasons:

1. **Fail-fast would break the default invocation.** `--height 480` is the CLI default for every command, and 480 isn't a multiple of 64 — so strict validation would make a bare `ltx-2-mlx ic-lora` error out. Fixing that would mean changing the default dims too, which is a user-visible behavior change I don't want to couple to this.

2. **Rounding is already the established behavior across the repo, not something #68 introduced.** `generate --two-stage` / keyframe have always encoded conditioning at `H_half * 32` derived from the latent grid (see `ti2vid_two_stages.py` — `enc_h_half = H_half * 32`), and even pre-#68 a plain ic-lora run at 480 already produced 448-high output (the latent grid floors `//32` regardless). Your fix just brought ic-lora's image-conditioning encode in line with what the rest of the two-stage pipelines already did. Fail-fast here would make ic-lora stricter than every sibling pipeline.

3. **This isn't the silent-wrong-result class that warrants hard failure.** Where we do fail loud in this repo (missing upsampler weights, missing `--dev-transformer`), silence meant *incorrect output*. Rounding produces a correct result at slightly different dims — the right loudness is a warning, not an error.

Concretely: when `height % 64 or width % 64`, print something like
`two-stage half-res snaps dims to multiples of 64; output will be 448x704 (requested 480x704). Use --single-stage for exact dims.`
Ideally as a small shared helper so `generate --two-stage`, keyframe, and ic-lora all report it consistently (they're all silent about it today).

Happy to take that in a follow-up PR if you want to open the issue — or feel free to send the patch yourself.

plz12345:
OK, cool, we're on the same page.  single stage would be divisible by 32 instead of 64, but agree on the shared helper implementation and stdout for the dimension snapping. I'll do a PR for it.

PR #72, opened 2026-07-11
feat: warn on output dimension snapping via shared helper

plz12345:
Follow-up to the dimension-snapping discussion on #68.

## What

Adds a shared `snap_output_dimensions(height, width, *, two_stage)` helper in
`ltx-core-mlx` (next to `compute_video_latent_shape`) and wires it into every
`generate` path. Two-stage pipelines generate Stage 1 at half resolution and
upscale 2x, so full dims must be multiples of 64; single-stage/one-stage need
multiples of 32. The helper is a pure, idempotent floor that clamps to at least
one tile and logs a **one-line informational warning** naming the resulting
output size — but only when a dim actually changes.

Per your call on #68, this is **snap-down + warn, not fail-fast**: rounding is
already established repo-wide behavior (non-conforming dims were floored
silently downstream — the CLI default `--height 480` isn't a multiple of 64, so
a bare two-stage run has always snapped 480→448), and it produces a correct
result at slightly different dims, so a warning is the right loudness. **Defaults
are unchanged.**

## Changes

- New `snap_output_dimensions` + module constant `SPATIAL_COMPRESSION` in
  `components/patchifiers.py`.
- Wired into the top of every generate path: one-stage, two-stage, two-stage-HQ,
  distilled, keyframe, a2vid, lipdub, ic-lora (`hdr_ic_lora` inherits ic-lora's).
  Callers pass `two_stage=` per topology; the snapped dims flow downstream so the
  output matches the warning.
- Collapsed ic-lora's existing explicit `(height//2//32)*32` half-res floor from
  #68 into the shared helper, so the invariant lives in exactly one place.
- Unit tests: both moduli, positive/negative paths, clamp-to-one-tile,
  idempotency, and warning content (including that the `--single-stage` hint is
  two-stage only).

## Local validation

M-series 64 GB, q8 dev transformer, 25 frames, `--frame-rate 24`, seed 42.
ffprobe reports `width,height` (W×H). Positive path = snap fires; negative =
clean dims pass through untouched. Both moduli covered.

| pipeline | requested (W×H) | snap warning | output (W×H) |
|---|---|---|---|
| `--one-stage` | 704×480 | none (÷32 clean) | **704×480** |
| `--one-stage` | 704×500 | `single-stage snaps dims to multiples of 32; output will be 480x704 (requested 500x704).` | **704×480** |
| `--two-stage` | 704×480 | `two-stage half-res snaps dims to multiples of 64; output will be 448x704 (requested 480x704). Use --single-stage for exact dims.` | **704×448** |
| `--two-stage` | 704×448 | none (÷64 clean) | **704×448** |

Two-stage default (704×480 requested) → **704×448**, byte-identical dims to the
explicit 704×448 run, confirming the snap only surfaces flooring that already
happened silently.

`ruff check` + `format --check` clean; full suite 594 passed / 22 skipped.

### DIFF
--- a/packages/ltx-core-mlx/src/ltx_core_mlx/components/patchifiers.py
+++ b/packages/ltx-core-mlx/src/ltx_core_mlx/components/patchifiers.py
@@ -5,7 +5,14 @@ Ported from ltx-core/src/ltx_core/components/patchifiers.py
 
 from __future__ import annotations
 
+import logging
+
 import mlx.core as mx
+
+logger = logging.getLogger(__name__)
+
+# VAE spatial compression: pixel dims must be multiples of this to encode cleanly.
+SPATIAL_COMPRESSION = 32
 
 
 class VideoLatentPatchifier:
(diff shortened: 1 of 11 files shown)


dgrauet:
## Review — code analysis + bit-exact e2e validation

Exactly what we agreed on in the #68 thread, cleanly executed. Verified in depth:

### Code review

- **The #68 floor replacement is mathematically exact**: `(h//2//32)*32 == ((h//64)*64)//2` by the nested floor-division identity (`floor(floor(h/2)/32) == floor(h/64)`), so collapsing ic-lora's explicit half-res floor into the shared helper changes nothing.
- **Upfront snap is latent-neutral everywhere**: every wired pipeline already floored its latent grid `//32` downstream, so snapping `height/width` first produces the same latent shape and the same conditioning encode dims (`H_half * 32`). It additionally hardens image-conditioning encodes at exotic dims on the paths that fed raw pixel dims to the VAE (same crash class #68 fixed for ic-lora).
- **Coverage is right**: all 8 generate paths wired (hdr-ic-lora inherits ic-lora's), and retake/extend correctly left out — their dims come from the source video, not a user request. `--upsample-only`/`--skip-stage-2` fall under `two_stage=not single_stage`, which is the correct modulus for both (half-res gen either way).
- **Clamp-to-one-tile** turns a previously-crashing corner (dims < 64) into a working floor — good.
- Tests cover both moduli, positive/negative paths, idempotency, clamp, and warning content. Nice — this also closes the "no new unit tests" nit from the #68 review.

### E2E (M2 Pro 32 GB, q8 `--low-ram`, seed 42, Union Control + synthetic canny — same rig/assets as the #68 validation)

| Check | Result |
|---|---|
| ic-lora 384×640×25 on this branch vs the #68-validation run (mult-64, snap = identity) | **SHA-256 identical** (`54f70aa…`) — bit-exact across both refactors |
| ic-lora 480×704×9 `main` vs this branch (snap 480→448 fires) | **SHA-256 identical** (`bad8063…`) — the warning only surfaces flooring that already happened |
| Warning behavior | exact message emitted on the 480 run; **silent** on clean dims |

So: defaults unchanged, outputs bit-for-bit unchanged, and users finally get told what resolution they're actually getting.

One cosmetic note, take it or leave it: the message prints `448x704` in H×W order (consistent with `--height`/`--width`), while ffprobe and common usage are W×H. Fine as-is since it matches the CLI flag order — just flagging the ambiguity.

LGTM.

dgrauet:
Follow-up on the cosmetic note — after checking conventions, I'd like the order flipped to **W×H before merge** rather than leaving it.

Rationale: video resolutions are universally quoted W×H (1920×1080, ffprobe, ffmpeg), and the repo's only explicit user-facing convention agrees — the `slice` command documents `--res` as "Target resolution WxH" (`cli.py:592`). The existing H×W in `iclora_utils.py:117`'s error message is the outlier, not the rule. The practical failure mode is when both dims differ: the warning would say `output will be 896x448` while ffprobe reports `448x896`, which reads like a rotated/wrong output.

Concretely:
- swap the args in the `logger.warning` call so it prints `{width}x{height}` for both the output and requested sizes (e.g. `output will be 704x448 (requested 704x480)`),
- adjust the two message-content test assertions accordingly.

Optional while you're in there: flipping `iclora_utils.py:117` to W×H too would remove the last H×W outlier — fine as part of this PR or not at all.

Everything else stands as reviewed (LGTM once flipped) — the e2e SHAs obviously aren't affected by a log-message change.

plz12345:
All set

Issue #74, opened 2026-07-14
Pipelines keep the DiT loaded across VAE decode — silent SIGKILL (jetsam) on 32 GB Macs


dgrauet:
## Observed (2026-07-14, M2 Pro 32 GB, macOS 26.5.2, MLX 0.31.1)

Two consecutive `--two-stage --low-ram` runs (704×1280, 97f, q8, LoRA strength 1.0) completed the full 89-minute denoising then died **silently** (no traceback, no crash report, no JetsamEvent) during `[Decoding video + audio + muxing]`, leaving a ~67 KB partial mp4.

smeltr memory polls (1 Hz, up to the last second before the kill) show the process entering decode holding **7.8 GB active + 17.65 GB MLX cache = 25.5 GB**, then getting SIGKILLed. A standalone decode of a same-shape random latent `(1,128,13,40,22)` from a fresh process peaks at **13.6 GB** and succeeds — the decode itself is fine; the killer is the baseline it starts from.

## Root cause

At LoRA strength 1.0 the stage-2 swap loads the **pre-fused `transformer-distilled` fully (non-streamed, ~8 GB q8)** and nothing frees it before `_decode_and_save_video`. 8 GB (DiT) + ~13.6 GB (decode peak) + system > 32 GB → jetsam. `LTX2_VAE_DECODE_BUDGET_GB=2` does not help because the budget only caps the block-3 activation, not the baseline.

## Fix (validated locally)

In `_base._decode_and_save_video`, when `low_memory`:

```python
if self.low_memory and self.dit is not None:
    self.dit = None
    self._loaded = False
    aggressive_cleanup()
```

before calling the orchestration decode. With this patch (currently applied uncommitted in the local working tree) the `--distilled` run decoded in 88.8 s and produced a valid mp4. `load()` reloads the DiT on demand for subsequent generations.

Needs: proper TDD-ified PR + a check that every pipeline (`ti2vid_two_stages`, `distilled`, `one_stage`, `ic_lora`, `keyframe`, `lipdub`) benefits since they all route through `_decode_and_save_video`.

## Side observation (upstream MLX?)

`mx.set_cache_limit(0)` is set by the low-ram path, yet `mx.get_cache_memory()` reads 3–17 GB during the run (0 in a standalone test). Suspicion: buffers freed while command buffers are in flight are counted as cache and not returned to the OS. Worth a minimal repro + upstream issue on ml-explore/mlx.

PR #80, opened 2026-07-19
fix: stop the Gemma encoder from widening a stricter cache limit (#79)

### LINKED ISSUE #79: low-ram: Gemma encoder clobbers mx.set_cache_limit(0) — cache grows to 8-21 GB during denoise/decode
## Observed

With `--low-ram`, `BasePipeline.__init__` sets `mx.set_cache_limit(0)` (`_base.py:82`) so freed buffers return to the OS. Yet on real runs (704×1280×97f distilled, M2 Pro 32 GB, smeltr 1 Hz memory polls) `mx.get_cache_memory()` reads **3-17 GB during denoise and 8-21 GB during video decode**, dropping to 0 only at process end. First noted as a side observation on #74 with an upstream-MLX suspicion.

## Root cause — not MLX

Six standalone repro scenarios (per-step eval pipelines, refs dropped while CBs are in flight via `mx.async_eval`, 48-op lazy chains polled from a side thread during `mx.eval`, conv2d stacks, `mx.compile` blocks) all show MLX 0.31.1 honoring `cache_limit(0)` perfectly: peak cache 0.00 GB in every case.

The clobber is ours: `ltx_core_mlx/text_encoders/gemma/encoders/base_encoder.py:83` — `_ensure_metal_headroom()`, called at **every prompt encoding** (`get_all_hidden_states`), does:

```python
mem_limit = mx.device_info()["memory_size"]
mx.set_cache_limit(int(mem_limit * 0.9))   # ~30.9 GB on a 32 GB Mac
```

inside a silent `try/except`, and never restores. From prompt encoding onward the whole run (transformer load, denoise, decode) runs with a ~0.9×RAM cache limit — the GB-level cache readings are the allocator doing exactly what it was (re)told to do. Control repro: `set_cache_limit(0)` then the clobber then a decode-like loop → immediate 1.07 GB cache retention (and `set_cache_limit`'s return value confirms the previous limit was 0).

This also explains the #74 side note that a **standalone** decode reads cache 0: it never runs the text encoder.

## Impact

In low-ram mode, GBs of freed buffers are retained as cache through the most memory-critical phases — directly feeding the jetsam pressure #74 addressed.

## Fix direction

`_ensure_metal_headroom` was meant to *cap* the cache below the default, not to widen a stricter configured limit. `mx.set_cache_limit` returns the previous value, so the fix is a min-semantics one-liner: apply the 0.9×mem cap, and if the previous limit was already stricter, restore it. No behavior change for default (non-low-ram) runs.

https://claude.ai/code/session_01HNcW12C5eFE9hCL2EnMYmW

dgrauet:
## Problem (#79)

`--low-ram` sets `mx.set_cache_limit(0)` at pipeline init, but `_ensure_metal_headroom()` (called at **every prompt encoding**) unconditionally reset the process-global limit to 0.9× device memory and never restored it. Everything downstream — transformer load, denoise, VAE decode — then ran with a ~30 GB cache limit: 3-17 GB of retained cache during denoise and **8-21 GB during decode** observed on a 32 GB Mac, feeding exactly the jetsam pressure #74 addressed. Six standalone repros confirmed MLX honors `cache_limit(0)` on every suspected path — the clobber was ours, not upstream's.

## Fix

`_ensure_metal_headroom` was always meant as a **cap** below the default, never a raise. Min semantics via `set_cache_limit`'s returned previous value: apply the 0.9×mem cap; if the previous limit was stricter, restore it. One method, one call site.

## Scope

Code touched: text encoder only. Behavior affected: the whole run in `--low-ram` mode (the limit is process-global and the encoder runs first). Non-low-ram runs: strictly unchanged (default limit > cap → capped as before, pinned by test).

## Validation — no regression across pipelines

- TDD: `tests/test_metal_headroom.py` — stricter limit (0 and intermediate) preserved, looser limit still capped.
- **Full suite: 625 passed**; the 3 failures (`test_generate_tiny_video`, `test_generate_small_video`, `test_t2v_real_prompt`) are pre-existing ffmpeg-environment failures, verified identical on clean `main`.
- Real `--distilled --low-ram` run under smeltr: **`cache_bytes` = 0 on all 276 memory polls across the entire run** (pre-fix equivalent run: up to 21.3 GB during decode), wall time unchanged (138.4 s vs 137.3 s — the encode now also runs at limit 0 in low-ram, and the alloc churn is free in practice).

Closes #79

https://claude.ai/code/session_01HNcW12C5eFE9hCL2EnMYmW

### DIFF
--- a/packages/ltx-core-mlx/src/ltx_core_mlx/text_encoders/gemma/encoders/base_encoder.py
+++ b/packages/ltx-core-mlx/src/ltx_core_mlx/text_encoders/gemma/encoders/base_encoder.py
@@ -77,10 +77,21 @@ class GemmaLanguageModel(nn.Module):
 
     @staticmethod
     def _ensure_metal_headroom() -> None:
-        """Set Metal cache limit to leave headroom for the GPU watchdog."""
+        """Cap the Metal cache limit to leave headroom (min semantics).
+
+        This is a CAP below the default, never a raise: a stricter limit
+        already configured (e.g. ``mx.set_cache_limit(0)`` by low-ram
+        streaming) is a deliberate setting and must survive prompt encoding
+        (#79 — the unconditional set used to clobber low-ram's 0 for the
+        whole rest of the run, retaining 8-21 GB of cache through decode).
+        ``mx.set_cache_limit`` returns the previous value, which is the only
+        way to read the current limit.
+        """
         try:
-            mem_limit = mx.device_info()["memory_size"]
-            mx.set_cache_limit(int(mem_limit * 0.9))
+            cap = int(mx.device_info()["memory_size"] * 0.9)
+            previous = mx.set_cache_limit(cap)
+            if previous < cap:
+                mx.set_cache_limit(previous)
         except Exception:
             pass
 
--- a/tests/test_metal_headroom.py
+++ b/tests/test_metal_headroom.py
@@ -1,0 +1,55 @@
+"""#79: the Gemma encoder must not widen a stricter configured cache limit.
+
+`--low-ram` sets `mx.set_cache_limit(0)` at pipeline init so freed buffers
+return to the OS. `_ensure_metal_headroom()` (called at every prompt
+encoding) used to unconditionally set the limit to 0.9x device memory,
+silently clobbering the low-ram setting for the whole rest of the run —
+observed as 8-21 GB of retained cache during video decode on 32 GB Macs.
+
+The headroom call was meant to CAP the cache below the default, never to
+raise a stricter limit: min semantics.
+"""
+
+from __future__ import annotations
+
+import mlx.core as mx
+import pytest
+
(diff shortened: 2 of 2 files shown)


dgrauet:
Additional validation — LoRA under block streaming (`--low-ram`), the heaviest allocator-churn path (per-block `BlockLoraSource` bind-time fusion, strength 0.8 to force it):

- `generate --distilled --low-ram --lora ltx-2.3-22b-distilled-lora-384.safetensors 0.8` (704×480×33f, seed 42) under smeltr: completed in 244.1 s (28.7 s/it — the expected bind-time fusion overhead vs 18.5 s/it plain), valid mp4 (33 frames + AAC).
- **`cache_bytes` = 0 on all 485 memory polls** across the whole run; `active` peaked at 13.96 GB and returned to 0 at exit.

The cache-limit-0 behavior now holds through the streaming+LoRA path too.

Issue #40, opened 2026-06-08
Q4 distilled → mosaic/rainbow-grid on M2/M3 GPUs (clean on M4 Max, identical MLX 0.31.1 / v0.14.8)


mrbizarro:
Hi @dgrauet — downstream maintainer here (Phosphene, the Pinokio panel that vendors ltx-2-mlx, currently pinned to **v0.14.8**). Flagging a consistent, hardware-correlated pattern in case it rings a bell or you can repro.

## Summary
The **Q4 distilled** model produces a **mosaic / rainbow-grid** pattern over most of the frame on **Apple M2 and M3** GPUs, while the **same model + same MLX render cleanly on M4 Max**. The render completes normally — both denoise stages and the VAE decode finish, and a full-size output file is written — so it's not a crash, OOM, or NaN blow-up; the pixels are just numerically garbage. At least one reporter has **Q8/High rendering clean while Q4/Quick garbles on the same machine**.

The clustering by **GPU generation** (rather than random across users) is what points us at a compute/precision path rather than a download/file problem — random file corruption wouldn't correlate with the chip.

## Environment matrix
| Reporter | Chip | RAM | macOS | MLX | ltx-2-mlx | Q4 | Q8 |
|---|---|---|---|---|---|---|---|
| us | M4 Max | 64 GB | 26.4 | 0.31.1 | v0.14.8 | **clean** | clean |
| poppy0396 | M3 Ultra | — | 26.5 | 0.31.1 | 0.14.8 | **mosaic (Quick)** | **clean (High)** |
| elbarto | M3 Max | 36 GB | 26.5.1 | 0.31.1 | 0.14.8 | **mosaic (all settings)** | — |
| rathore | M2 | 16 GB | — | — | 0.14.8 | **mosaic (Quick)** | — |

- M4 Max renders clean even with a minimal 6-file Q4 install.
- Affected users pass a header+size integrity check on all Q4 files. (We've shipped a SHA-256-vs-HF-upstream content verifier downstream and are collecting results to fully rule out content corruption — will update here — but the chip-generation clustering already argues against that.)
- Reproduces across Quick / Balanced / Standard — i.e. all the Q4 `DistilledPipeline.generate_two_stage` paths.

## Representative log (M3 Ultra, I2V Balanced 1024×576)
```
[Loading transformer (transformer-distilled.safetensors)] done in 0.5s
[...]

Issue #34, opened 2026-06-04
Near-silent audio on mlx 0.31.2: strided .at[].add() scatter mis-indexes source (UpSample1d / HannSincResampler)


dgrauet:
## Symptom

Generated videos have **near-silent audio** (mean ≈ −52 dB, max ≈ −37 dB) on **mlx 0.31.2**. The same code on **mlx 0.31.1** produces normal audio (≈ −13 dB / −3 dB). Video is unaffected. First observed as the "audio fade / quasi-silent audio" regression during ComfyUI-LTXVideo-mlx end-to-end testing (2026-05-05).

## Root cause

**MLX 0.31.2 regression**: `mx.array.at[<strided slice>].add(x)` mis-indexes the *source* on Metal — element `x[k]` is written as if indexed by the *destination* row (`y[2k] ← x[2k]` instead of `y[2k] ← x[k]`). Plain strided slice *assignment* (`y[:, ::2, :] = x`) is correct; only the `.at[].add()` scatter path is broken. Small shapes are unaffected (e.g. `(1, 16, 1)` is correct, `(2, 128, 8)` is wrong), so a vectorized kernel path is the likely culprit.

Minimal repro (no LTX code):

```python
import mlx.core as mx
import numpy as np

B, T, C = 2, 64, 8
x = (mx.arange(B*T*C, dtype=mx.float32).reshape(B, T, C) % 7 - 3) / 3.0
y = mx.zeros((B, T*2, C)).at[:, ::2, :].add(x)
mx.eval(y)

expected = np.zeros((B, T*2, C), dtype=np.float32)
expected[:, ::2, :] = np.array(x)
print("wrong elements:", int((np.array(y) != expected).sum()))
# mlx 0.31.1 -> 0      mlx 0.31.2 -> 872
```

## Affected call sites (both in the audio path)

- `packages/ltx-core-mlx/src/ltx_core_mlx/model/audio_vae/vocoder.py:125` — `UpSample1d` (anti-aliased Activation1d upsample): `x_up.at[:, ::2, :].add(x)`
- `packages/ltx-core-mlx/src/ltx_core_mlx/model/audio_vae/bwe.py:95` — `HannSincResampler`: `upsampled.at[:, ::ratio].add(x_padded)`

The corrupted zero-insert feeds every `SnakeBeta` activation in BigVGAN (≈108 convs), collapsing the waveform to noise ≈ −50 dB.

## Evidence chain (full bisection)

[...]

PR #83, opened 2026-07-26
perf(media): zero-copy frame/audio writes, opt-in write overlap

nuclearspike:
**Summary of the revision:** end to end this change measures no reliable
difference (0.8% on your A/B, inside noise). What it does reliably reduce is
transient allocation on the write path, which is unconditional and free. The
write-overlap thread, which is the part that could bite, is now off by default.

## What changed

Two independent changes to the media write path:

1. **Zero-copy writes (unconditional).** Contiguous MLX RGB frame buffers go
   straight to ffmpeg through `memoryview`, and PCM audio is quantized and
   interleaved in MLX and written from its own contiguous buffer. Neither
   materializes a Python `bytes` copy. Short writes are handled in a loop
   rather than falling back to a full copy.
2. **`_OrderedFrameWriter`, an optional write-overlap thread (opt-in, default
   off).** When `LTX2_MEDIA_WRITE_OVERLAP=1`, a single-worker
   `ThreadPoolExecutor` keeps at most one write in flight so the write can
   overlap the next frame's decode. Ordering and byte output are unchanged;
   with the variable unset there is no extra thread and no concurrency added
   to the decode loop.

Also: byte-exact mono/stereo and partial-write regression tests for the audio
and frame write paths.

## Why

The previous path materialized a full `bytes` object per video frame and copied
the whole waveform through NumPy before writing, even though producer and
consumer both accept contiguous buffers. At HD resolutions those transient
copies add allocator pressure during the VAE/mux phase.

To be explicit about what this is not: the pre-existing code was **correct**,
just costly — `bytes(memoryview(strided))` goes through `tobytes`, which is
stride-aware. This is a performance change, not a bug fix.

## Measured impact

**End to end: no measurable change.** On a 512×512×25 q8 distilled render
(`--low-ram`, audio included, fixed seed), overlap on vs off vs `main` differ by
0.8% wall clock — inside run-to-run noise — and all three outputs are
byte-identical. That measurement is the maintainer's, reproduced from the review
of this PR, and it is the number that matters for the decision below.

The local write loop does improve, on a 1280×704×121 serialization
microbenchmark:

- write-loop median: 4.431 ms → 0.078 ms
- peak transient Python allocation: 2,703,933 bytes → 512 bytes

With overlap enabled, that write-loop figure reflects work **moved onto the
worker thread**, not work eliminated; the zero-copy portion is a genuine
reduction in allocation either way. Denoise and VAE decode dominate full-render
time, so no end-to-end speedup is claimed.

**Because the overlap thread buys nothing measurable at these sizes, it now
defaults to off** — the free, unconditional zero-copy win stays on, and the
concurrency is opt-in until a workload demonstrably benefits.

## The property this design depends on

Handing a `memoryview` of an MLX array to a worker (or to `stream.write`) while
the owning reference is dropped and `aggressive_cleanup()` runs relies on MLX
keeping the exporting array alive for the life of the buffer export. That is an
implementation property of MLX, not something this code can enforce, and if a
future release stopped honouring it the failure would be silently corrupted
frames rather than an exception.

You probed that directly (20/20 survival across `del` + `clear_cache()`, 40/40
under concurrent read) and wrote `tests/test_media_buffer_lifetime.py` to pin
it. I've asked for that file to land on this branch before merge, so the
invariant is covered by the same commit range that depends on it.

## Behavior and interface changes

- New environment variable `LTX2_MEDIA_WRITE_OVERLAP` (default `0`; `1`/`true`/
  `yes`/`on` enable). No other configuration changes.
- `_write_all` raises `OSError` naming the byte counts when a stream reports a
  non-positive write, instead of `BrokenPipeError("stream accepted no data")` —
  a zero-length write from a `BufferedWriter` is a stalled stream, not a closed
  pipe, and the old exception would have misled whoever read the traceback.
- ffmpeg teardown (`stdin.close()`, `proc.wait()`, `aggressive_cleanup()`) moved
  inside the `finally`. It previously sat after the `try`, so only a
  `BrokenPipeError` reached it; any other decode-loop exception left ffmpeg
  running until garbage collection. The new `OSError` path made that reachable,
  so the fix ships with it.
- No new dependencies. No public API changes.

## Validation

Re-verified for this revision on macOS 26.5.2, Apple M5 Max, Python 3.12.11,
MLX 0.32.0:

- `pytest tests/test_media_serialization.py` — 15 passed (the 6 existing media
  tests plus the parsing and error-type pins added here)
- `pytest tests/test_trainer_core.py` — 32 passed, as a check that the teardown
  change is inert outside the decode path
- `ruff format --check` and `ruff check` on `packages/` and `tests/` — clean
- both new tests were confirmed to FAIL against the pre-fix behaviour before
  landing (default restored to on; `OSError` restored to `BrokenPipeError`)

I did not get a clean full-suite run locally: on this machine the suite dies
part-way through `tests/test_generate.py` and, with the generation tests
excluded, in `tests/test_trainer_core.py` — yet that same file passes 32/32 when
run alone, so it is whole-suite resource exhaustion here, not a failure this
change introduces. CI on the refreshed base is the authority for the full suite: it had never run
on this PR until you refreshed the base and approved the workflow on 16 August,
and it was 5/5 green on `31a5213` once it did. It will re-run on this revision.

For the record, the original submission reported `pytest -m 'not slow'` at 615
passed / 1 skipped, with three pre-existing failures in ffmpeg tests that close
`stdin` and then call `communicate()`, raising `ValueError: flush of closed
file` on Python 3.11. That diagnosis stands and is unrelated to this change; I
have not re-run that configuration here.

Byte-identical output is attested twice, independently:

- your A/B on M2 Pro / MLX 0.31.1 — three renders (`main`, overlap-on,
  overlap-off) byte-identical, audio present at −36.2 dB mean
- the original submission's 1280×704×129 BF16 LTX-2.3 render, MP4 SHA-256
  `247c3bbdd23794a72263f9229e2196780443c29b66cd24f693085ded8db91f03`
  matching baseline (produced on the earlier MLX 0.31.x runtime, not re-run for
  this revision)

Not tested here: CUDA/Linux (this is an Apple-silicon MLX path) and Python 3.11.


nuclearspike:
Production validation update (Apple M5 Max, 128 GB, MLX 0.31.1):

- The integrated zero-copy/bounded media path ran in the exact clean runtime `3171bac4ba901c0237faea2678c34034b37abc2a` (`0.14.20.dev1`).
- Fixed-seed BF16 1280x704x129 AUTO→eager output is byte-identical to the eager control, including the final muxed MP4 SHA-256.
- Un-tiled low-RAM remains byte-identical while reducing MLX peak from 46.140 GiB to 17.123 GiB.
- Media-focused tests, ruff, and the full integrated pure/synthetic tests pass.
- Temporal/spatial modality tiling was explicitly rejected on quality; no quality claim is being conflated with these media-copy changes.

This PR is ready for maintainer review.

dgrauet:
Reviewed, tested, and validated end-to-end. This is good work and I'd
merge it. One request at the end, plus a suggestion about how the PR
describes itself.

Note first that CI had never run on this PR — as a fork PR it sat in
`action_required` since 26 July, so the green you may have expected was
never there. I refreshed the base onto current `main` and approved the
run; your `31a5213 style: apply ruff format` cleared the last red. It is
5/5 green now.

## The PR does more than its title says

"stream MLX buffers without copies" describes half of it. The other half
is `_OrderedFrameWriter`, a `ThreadPoolExecutor` doing **asynchronous
writes, on by default** via `LTX2_MEDIA_WRITE_OVERLAP`. That is the
structurally significant change here and it isn't mentioned in the title
or the summary. Worth saying explicitly so the git history records what
actually landed.

It also reframes the headline number. "write-loop median 4.431 ms →
0.078 ms (−98.2%)" measures a loop whose work **moved to another
thread**, not work that disappeared. You're upfront that there's no
broad end-to-end claim, which I appreciate — but see the timings below.

## Validation

Three renders, same seed, same prompt, q8 distilled 512×512×25,
`--low-ram`, audio included:

| run | branch | `LTX2_MEDIA_WRITE_OVERLAP` | wall clock | sha256 |
|---|---|---|---|---|
| A | `main` | — | 111.3 s | `07cfc7cf…` |
| B | this PR | `1` (default) | 110.4 s | `07cfc7cf…` |
| C | this PR | `0` | 111.7 s | `07cfc7cf…` |

**All three byte-identical.** That settles both questions at once: the
no-copy path doesn't change output, and neither does the writer thread.
Audio is real too (mean −36.2 dB, max −17.7 dB — not the near-silence of
#34).

The end-to-end delta is **0.8%**, inside measurement noise. So the thread
buys nothing measurable at this scale, which is worth knowing when
weighing it against the concurrency it introduces.

I also checked the risky property directly rather than assuming it. The
design hands a `memoryview` of an MLX array to a worker, then drops its
own reference and runs `aggressive_cleanup()` — `gc.collect()` plus
`mx.clear_cache()` — every eight frames while that write is in flight. On
M2 Pro / MLX 0.31.1:

- buffer survives `del` + `clear_cache()` + heavy reallocation: 20/20
- concurrent read while the owner churns and clears: 40/40

So MLX's buffer protocol does keep the exporting array alive, and the
mechanism is sound. `mx.contiguous` is correctly placed, too —
`memoryview(...).cast("B")` is only defined for C-contiguous views.

(Incidentally that confirms the pre-PR code was *correct*, just costly:
`bytes(memoryview(strided))` goes through `tobytes`, which is
stride-aware. So this is a pure performance change, not a bug fix.)

The audio path is clean and numerically identical — `np.clip` +
`astype(int16)` and `mx.clip` + `astype(mx.int16)` both truncate toward
zero.

## The request: pin the property the design rests on

All of that safety is an **implementation property of MLX**, not
something the calling code can enforce. If a future MLX release stopped
honouring buffer exports against `clear_cache()`, the result would be
silently corrupted video frames — no exception, no traceback. That is
the exact shape of #34 (near-silent audio from a mis-indexed scatter) and
#40 (M2/M3 mosaic), both of which cost this repo real time.

Your five tests exercise `_OrderedFrameWriter` in isolation — partial
writes, byte exactness, pipe-failure propagation — but none of them puts
`aggressive_cleanup()` between submissions with a write outstanding,
which is precisely the decode loop's shape.

I wrote the missing tests rather than just asking for them:
`tests/test_media_buffer_lifetime.py`, six tests, ~0.3 s, ruff-clean.
They cover the buffer surviving owner release plus cleanup, the same
under concurrent read, `_OrderedFrameWriter` with cleanup in flight
(parametrised over both overlap modes), overlap-on vs overlap-off
byte equality, and one that pins `mx.contiguous` by asserting the
`TypeError` you get without it — so a future "cleanup" that removes the
call is caught here.

Say the word and I'll push the file to this branch, or open it as a
follow-up PR against `main` if you'd rather keep yours untouched.

## Two nits

- `_write_all` raises `BrokenPipeError("stream accepted no data")` when
  `written <= 0`. A zero-length write from a `BufferedWriter` isn't a
  broken pipe; the exception misstates the cause. Harmless in practice,
  but it would mislead whoever reads that traceback.
- Given the 0.8% end-to-end delta, consider whether
  `LTX2_MEDIA_WRITE_OVERLAP` should default to `0` — keeping the no-copy
  win, which is free and unconditional, and making the thread opt-in
  until there's a workload that demonstrably benefits. Your call; I'm
  not blocking on it, since the byte-identical results hold either way.

nuclearspike:
Thank you — the buffer-lifetime probes and the three-way byte-identical A/B are
more than I had, and both critiques land.

**Yes, please push `tests/test_media_buffer_lifetime.py` to this branch**, and
I'd rather it land before you merge: the property it pins is exactly the one
that fails silently, so the code and its test should enter `main` together
rather than across two PRs.

You were right that the title described half the diff. `_OrderedFrameWriter` was
the structurally significant part and it appeared in neither the title nor the
summary. The title now names the write overlap, and the body opens with the
honest result — your 0.8%, inside noise — before anything else, and says
explicitly that the write-loop figure reflects work moved onto the worker
thread, while the allocation reduction is real either way.

Both nits are fixed, and chasing the second one turned up a third problem:

- `LTX2_MEDIA_WRITE_OVERLAP` now defaults to `0`. Your reasoning is the deciding
  argument: the no-copy win is free and unconditional, the thread is neither,
  and 0.8% doesn't justify adding concurrency to the decode loop by default.
  `1`/`true`/`yes`/`on` enable it; anything else is off.
- `_write_all` now raises `OSError` naming the byte counts. A zero-byte write
  from a `BufferedWriter` is a stalled stream, not a closed pipe, and the old
  exception would have sent the next reader of that traceback the wrong way.
- **That change was not safe on its own.** `stdin.close()`, `proc.wait()`, and
  `aggressive_cleanup()` sat *after* the `try`, so only `BrokenPipeError`
  reached them; the new `OSError` propagated past and would have left ffmpeg
  running until GC. Teardown moved into the `finally`, which also closes the
  pre-existing hole for any other exception raised inside the decode loop.

Chasing that one turned up a fourth: `proc.stdin.close()` flushes buffered
bytes, so it raises `BrokenPipeError` when ffmpeg has already exited — which in
the new teardown block would have skipped `proc.wait()` and the cleanup on
exactly the path where the process had already failed. The close is now guarded.

One thing I noticed and deliberately did **not** change, since it predates this
PR and the fix is yours to choose: `stderr=subprocess.PIPE` is never read, so
`proc.wait()` can block if ffmpeg fills that pipe buffer. Moving teardown into
the `finally` means `wait()` now runs on exception paths that previously skipped
it, so this change widens the exposure without creating it. If you'd like a
timeout-and-kill or a stderr drain, say so and I'll add it here; otherwise I've
left your process handling alone.

Both new behaviours are pinned by tests that I confirmed fail against the
previous behaviour first: `LTX2_MEDIA_WRITE_OVERLAP` parsing across unset and
eight spellings, and the stalled-write error asserting the type is exactly
`OSError` — a bare `raises(OSError)` would have accepted a regression back to
`BrokenPipeError`, since it subclasses `OSError`.

On CI never having run: that one is on me. I read "no failures" as "green" for
three weeks without checking that a workflow had actually started. Thank you for
refreshing the base and approving the run.

dgrauet:
Everything is in place except one mechanical thing, and it's the only
reason this isn't merged yet.

**Done on my side:**

- `tests/test_media_buffer_lifetime.py` is pushed to this branch as
  `dc296da`, as you asked — code and test enter `main` together.
- Base refreshed onto current `main` (which now carries #88 and
  v0.14.21), and I approved the workflow run.

**Results:**

| check | result |
|---|---|
| `test (3.11)` / `test (3.12)` on `macos-14-arm64` | pass |
| `lint`, `syntax` | pass |
| `commitlint` | **fail** |

Worth calling out: the buffer-lifetime tests pass **on the M1 runners**,
not just my M2 Pro. That matters more than it might look — #86 in this
repo is currently blocked on numerics that hold on M4 Max and M2 Pro and
break on M1, so "it passed on my machine" has a poor track record here.
Your design holds across all three.

Locally: 21 tests across the two media files (your 15 + my 6), full suite
628 passed, ruff clean.

**The blocker: `commitlint` rejects the five revision commits.**

```
27e8bfa  Reap ffmpeg regardless of why closing stdin failed
44f2e9f  Make ffmpeg reaping structural rather than incidental
b0accab  Reap ffmpeg even when closing its stdin fails
16f1169  Keep ffmpeg teardown on any write error, and pin both behaviours
f229d58  Make write overlap opt-in and correct the stalled-write error
```

Each fails `type-empty` / `subject-empty` — no conventional-commit type.
Your first three (`perf(media): …`, `style: …`) are fine; it's only these
five. `commitlint` is a required check, so the PR can't merge while it's
red.

Could you reword them? An interactive rebase over the five, or a squash
of them into one, whichever you prefer. Suggested types, purely as a
starting point:

```
fix(media): make write overlap opt-in and correct the stalled-write error
fix(media): keep ffmpeg teardown on any write error, and pin both behaviours
fix(media): reap ffmpeg even when closing its stdin fails
refactor(media): make ffmpeg reaping structural rather than incidental
fix(media): reap ffmpeg regardless of why closing stdin failed
```

Rebase freely — my test commit is at the tip and can be reordered,
squashed or dropped and re-pushed, whatever is convenient. Just don't
lose the file itself.

To be straight with you about why this is being asked at all: this repo
squash-merges, so none of these messages reach `main` — what lands is the
squash subject, taken from the PR title, which is already conventional.
So the check is enforcing something with no effect on the history it
guards. That's a misconfiguration on our side, not a failing on yours,
and I've flagged it separately. But it's a required check today, so it
has to go green before this can land.

Once it's green I'll merge. Nothing else is outstanding.

On the `stderr=subprocess.PIPE` point: agreed it's pre-existing and I'd
rather not fold it into this PR. I'll open an issue for it so it doesn't
get lost — your teardown change makes `wait()` reachable on paths that
previously skipped it, which is worth writing down while the reasoning is
fresh.

nuclearspike:
Reworded and pushed as `a9f2626`, rebased onto current `main`. I used your
suggested types verbatim; your test commit is replayed unchanged at the tip with
its authorship intact.

```
test(media): pin the exported-buffer lifetime the no-copy path relies on
fix(media): reap ffmpeg regardless of why closing stdin failed
refactor(media): make ffmpeg reaping structural rather than incidental
fix(media): reap ffmpeg even when closing its stdin fails
fix(media): keep ffmpeg teardown on any write error, and pin both behaviours
fix(media): make write overlap opt-in and correct the stalled-write error
```

`git diff f822e78 a9f2626` is empty, so this changed commit messages and
nothing else. Locally:
`commitlint` passes on all nine messages, the two media files are 21 passed,
and `ruff format --check` / `ruff check` are clean.

It's sitting in `action_required` again — a fork PR needs a workflow approval
per push, so it needs one more click from you before the checks run.

Thank you for writing the buffer-lifetime tests rather than just asking for
them, and for the M1 result — that the property holds on all three chips is
more reassuring than my own machine could ever be, given what #86 is stuck on.

PR #85, opened 2026-07-28
chore(deps): bump mlx 0.31.1 -> 0.32.2

plz12345:
## What

Lockfile-only bump of `mlx` (and `mlx-metal`) from 0.31.1 to 0.32.0. No source changes.

The `pyproject.toml` constraint is already `mlx>=0.31.0`, so 0.32.0 was permitted — this just moves the dev/CI environment. Consumers of the published packages resolve independently and are unaffected.

## Why

The parts of [MLX 0.32.0](https://github.com/ml-explore/mlx/releases/tag/v0.32.0) that actually touch this repo:

- **`Fix qmm_naive K-tail dispatch for FP quantized kernels`** — the main reason to take this. The q4/q8 transformer path round-trips weights through `mx.dequantize` → LoRA fuse → re-quantize (`packages/ltx-core-mlx/src/ltx_core_mlx/loader/fuse_loras.py`), so quantized-matmul correctness is on the critical path for every quantized generation. A dispatch bug for K values with a tail is squarely in that territory.
- **`Fix data race in tracing state by making it thread-local`** — general robustness; no known symptom here today.
- **Complex VJPs for `log`/`exp`** — no effect currently. The audio VAE's `mx.fft.rfft` (`model/audio_vae/processor.py`) and conv1d-basis STFT (`model/audio_vae/bwe.py`) are inference-only. This matters only if a spectral/mel loss that backprops through a log-magnitude is added to `ltx-trainer`.
- **JIT-compiled CUDA `qmm_naive`/`qmm_sm80`/`qmm_sm90`/`gather_gemm`** — noted for completeness; no effect on the Metal path.

Not applicable, for the record: `gather_qmm` batch-stride fix (no MoE in the tree), `qmv_wide` small-batch quantized matvec (LLM-decode kernel; video diffusion runs long sequences through GEMM), the rope single-token fix, and the axis=0 `concatenate` fast path (the only axis=0 concat is in a per-batch padding branch in the Gemma embeddings connector).

The new ops in 0.32.0 (`mx.flip`, `mx.unstack`) would tidy a couple of spots — the `for b in range(B)` + `mx.stack` loop in `text_encoders/gemma/embeddings_connector.py`, and the `[::-1]` reflection padding in `model/audio_vae/processor.py` — but they're readability-only with no perf benefit, so they're deliberately left out to keep this to one concern.

## Testing

Full suite on this branch: **607 passed, 22 skipped**. (The `slow` marker is declared in `pyproject.toml` but currently carried by no tests, so this is the whole suite.)

**Not validated against generation output.** Confirming the quantized path is numerically unchanged would need a q8 run (e.g. the 512×512×25 distilled smoke config) diffed against a pre-0.32 baseline. Happy to do that before merge if you'd rather not take it on unit tests alone.

## Drive-by: CI lint fix

A second commit reformats three Python snippets in `CLAUDE.md`. Unrelated to
the bump, but it's what was reddening `lint` here.

ruff 0.16.0 formats Python code blocks embedded in Markdown, and the lint job's
`ruff-action` appends the repo root to the paths it checks — so `CLAUDE.md` is
now in scope of `ruff format --check`. This is version drift, not something
this branch introduced: `main` was last green on 2026-07-21, before 0.16.0, and
every PR opened against it now fails the same way.

Docs-only, semantically neutral (blank line after an import, `i:i+1` →
`i : i + 1`, de-aligned trailing comments). `ruff 0.16.0 check` and
`format --check` over `packages/ tests/ .` are both clean on this branch.

Happy to split it into its own PR if you'd rather land it independently — in
that case this branch rebases onto it and drops the commit.


plz12345:
Heads up: the `lint` failure on this PR is unrelated to the bump — this branch changes `uv.lock` only, no `.py` or `.md` files.

`ruff-action` resolves ruff from the `>=0.7.0` spec in `pyproject.toml`, so it floats to whatever is newest at run time:

- main's last green CI run (#81, Jul 21) installed **ruff 0.15.22**
- this PR's run installed **ruff 0.16.0**

0.16.0 formats Markdown by default, where 0.15.x refused to without preview mode. Reproduced locally on an unmodified `CLAUDE.md`:

```
$ uvx ruff@0.15.22 format --check CLAUDE.md
error: Failed to format CLAUDE.md: Markdown formatting is experimental, enable preview mode.

$ uvx ruff@0.16.0 format --check CLAUDE.md
1 file would be reformatted
```

So `main` will go red on its next push too, independent of this PR. The fix is either pinning ruff (`ruff>=0.7.0,<0.16`) or running `ruff format CLAUDE.md` once and taking the reformat — a separate concern from this bump, so I've left it out. Happy to send that as its own PR if useful.

dgrauet:
Took you up on the validation you offered, and it turned into rather more
than a spot check — thank you for flagging that unit tests alone were not
enough here, because they were not.

## The stated reason for the bump does not apply to us

The PR leads with the `qmm_naive` K-tail fix, on the grounds that
quantized matmul is on the critical path of every quantized generation.
That reasoning is sound, but the premise is wrong: the fix targets **FP**
quantized kernels, and this repo's packs are affine int8/int4.

Fingerprinted `quantized_matmul` on six real DiT shapes, affine int8 and
int4 at group 64:

```
DIGEST fb1605008104709ee282dce8457c2ddd   ← 0.31.1
DIGEST fb1605008104709ee282dce8457c2ddd   ← 0.32.0
DIGEST fb1605008104709ee282dce8457c2ddd   ← 0.32.1
```

Bit-identical. So that particular fix buys us nothing — measured, not
argued.

## But the output changes anyway

Same seed, q8 distilled 512×512×25:

| mlx | sha256 | PSNR vs 0.31.1 |
|---|---|---|
| 0.31.1 | `07cfc7cf…` | — |
| 0.32.0 | `c4eec4ab…` | 29.56 dB |
| 0.32.1 | `3c9c1524…` | 29.56 dB |

Each version is deterministic across repeat runs. 0.32.0 and 0.32.1 are
close to each other (44.6 dB), so the shift happened in 0.32.0 and 0.32.1
only nudges it.

Bisecting by layer puts the origin in the DiT forward: text embeddings are
identical across all three, quantized matmul is identical, latents differ.
Best guesses from the 0.32.1 notes are `Round MLX_SDPA_BLOCKS up to a
multiple of 32` or the NAX attention `unroll_count(4)` change, but I have
not proven which.

## No regression, though — which is what actually matters

Bumping mlx is normal lifecycle; freezing it forever is not a strategy.
The question is whether each bump regresses anything, so I built a
harness to answer that repeatably — merged as #98,
`scripts/mlx_upgrade_report.py`. On 0.31.1 → 0.32.1, M2 Pro:

```
layer1_quantized_matmul      IDENTICAL
layer2_text_embeddings       IDENTICAL
layer3_latents               CHANGED
layer4_eager                 CHANGED   111.0s -> 110.2s (0.99x)
layer4_low_ram               CHANGED   102.1s ->  99.5s (0.97x)
peak GPU memory              15.25 GB -> 15.26 GB
```

Marginally faster, memory flat, 667 unit tests pass on both. Nothing that
counts as a regression. The output differs, but a hash moving is not a
regression — kernel selection and reduction order are allowed to change,
and gating on bit-identity would just train us to ignore a red light.

One observation worth recording rather than acting on: on 0.31.1 the eager
and `--low-ram` renders happen to produce the same bytes; on 0.32.x they
diverge. That is not a broken invariant — CLAUDE.md already documents the
compiled block differing from eager by ~1 fp32 ULP, and
`test_wrapper_matches_baseline` still passes at `allclose(1e-5)`. The
coincidence on 0.31.1 was luck at this configuration, and chaotic
amplification over 11 denoise steps is enough to break it.

## Suggestion

**Retarget this to 0.32.1** rather than 0.32.0. It supersedes it, ships the
`captured random state in compile` fix (relevant to `--low-ram`, which
compiles its block), and is the same distance from 0.31.1 on every axis I
measured. I have a branch with the lockfile bump ready if it is easier for
me to push it here — say the word, or bump it yourself, whichever you
prefer.

Whichever version we land, the CHANGELOG entry should say that seeds do not
reproduce across the upgrade. That is the real user-visible cost, and it is
worth stating plainly rather than discovering downstream.

Your drive-by CLAUDE.md fix is already in via #89, so this branch is now
`uv.lock` only after a base refresh — one clean concern.

dgrauet:
Retargeted this onto current main and **mlx 0.32.2** (0.32.0 → 0.32.1 → 0.32.2 released since this was opened) — thanks for kicking it off; your authorship is kept as a co-author on the commit.

Validation (layered regression report, `scripts/mlx_upgrade_report.py`, M2 Pro):

- **Layer 1 — `quantized_matmul` on the six real DiT shapes: bit-identical** between 0.31.1 and 0.32.2.
- First divergence is **layer 2 (Gemma forward)** — kernel-selection/reduction-order class, expected between releases; everything downstream follows from it.
- Renders quality-reviewed and approved (audio −35.7 dB, healthy); timings within noise (eager +5 %, `--low-ram` −2 %); peak memory unchanged.
- One behavioural note: eager and `--low-ram` renders are **no longer byte-identical** under 0.32.2 (`mx.compile` fusion moved). Both healthy; the existing ~1-ULP tolerance test covers it.
- Full suites green under 0.32.2: 802 non-slow + 35 slow, **including the 14 gemma4 torch-parity tests** (regenerated reference).

Merging on green CI.

PR #86, opened 2026-08-03
AdaLN: sigma-dedupe + lazy per-block gather — bitwise-identical, multi-GB peak-memory cut on Apple Silicon

mrbizarro:
## Motivation

On a unified-memory Mac, peak GPU allocation is the binding constraint long before FLOPs are. The q4 tier exists precisely so 16 GB and 32 GB machines can run this — and on those machines a couple of gigabytes of transient allocation is the difference between a render finishing and the Metal allocator refusing.

Whenever per-token timesteps are active — i2v, keyframe/FFLF, extend, retake, reference-latent conditioning, IC-LoRA, lipdub; everything except plain t2v — the per-token AdaLN path allocates far more than it needs to, in two separate ways. Both are addressed here, in two commits, with no change to a single output bit.

### 1. `perf(transformer): deduplicate the per-token AdaLN GEMM`

`LTXModel._adaln_per_token` runs the AdaLN MLP + projection over every one of the `B*N` tokens, every step. But the per-token conditioning vector only ever takes a handful of distinct values: the target sigma, plus one per conditioning group. Every conditioning path in the tree builds its per-token sigma as `denoise_mask * sigma` where the mask is assembled from `mx.ones` / `mx.zeros` / `mx.full(1 - strength)`, and `strength` is a scalar per conditioning item — so distinct rows number 2 to 4 in practice, not 20 000.

The flattened rows are grouped, the MLP runs on the distinct set, and the result is expanded back. Isolated cost of exactly the AdaLN work one denoise step does in per-token mode (video 9-param + 4-param at dim 4096 plus the audio pair at 2048), median of 5, M4 Max:

| video tokens | before | after | saved / step |
|---:|---:|---:|---:|
| 8 192 | 0.327 s | 0.055 s | 0.273 s |
| 14 080 | 0.566 s | 0.065 s | 0.502 s |
| 20 000 | 0.795 s | 0.074 s | 0.721 s |
| 30 000 | 1.196 s | 0.094 s | 1.102 s |

In context that is ~1 % of a denoise step — the AdaLN GEMM is 10.5 TFLOP of roughly 760 TFLOP at 20 000 tokens. Real, free, but nobody will feel it. It is here mainly because it is the prerequisite for the second commit.

### 2. `perf(transformer): defer the per-token AdaLN gather into the blocks`

This is the one that matters.

The projection's output is **float32** — the sinusoidal timestep embedding is float32 and MLX promotes — and it is shaped `(B*N, num_params*dim)`. At 20 000 video tokens that is **2.95 GB** for the 9-parameter video head plus **1.31 GB** for the 4-parameter AV cross-attention head, and it stays resident for the entire 48-block stack because every block needs it. Each block then materialises its own full-size `(B, N, P, dim)` copy in `_unpack_adaln` in order to slice the parameters out.

Since commit 1 already computes the distinct rows, keep the tensor in that form. `_adaln_per_token` returns a `PerTokenAdaLNParams` — the distinct rows plus the inverse index, a few kilobytes — and `BasicAVTransformerBlock._unpack_adaln` expands one `(B, N, dim)` parameter slice at a time, inside the block that consumes it, where MLX frees each slice as it goes. The output-block embedded timestep is expanded after the stack, at its single point of use.

Peak GPU memory for one denoise step, `mx.get_peak_memory()`, M4 Max 64 GB, q4 distilled with all 48 blocks resident, i2v-shaped per-token sigmas (2 distinct). The 10.54 GB of resident weights is included in every figure, and each number is the min of 2–3 measured steps after warm-up (they were identical to the megabyte across repeats):

| video tokens | current `main` | + commit 1 | + commit 2 | saved vs `main` |
|---:|---:|---:|---:|---:|
| 8 192 | 16.25 GB | 16.13 GB | **12.73 GB** | **3.52 GB** |
| 20 000 | 24.06 GB | 23.76 GB | **15.50 GB** | **8.56 GB** |

Step activations alone — peak minus the 10.55 GB weights-and-inputs baseline:

| video tokens | current `main` | + commit 2 | cut |
|---:|---:|---:|---:|
| 8 192 | 5.70 GB | 2.18 GB | −62 % |
| 20 000 | 13.51 GB | 4.95 GB | −63 % |

Wall clock is unchanged to slightly better, so this is not a memory-for-time trade:

| measurement | `main` | commit 1 | commit 2 |
|---|---:|---:|---:|
| synthetic step, 8 192 tokens | 26.80 s | 26.54 s | 26.21 s |
| synthetic step, 20 000 tokens | 88.10 s | 88.04 s | 86.37 s |
| real render, 1024×576×97 i2v, stage-2 step (7 488 tokens) | 24.52 s | 24.42 s | 24.17 s |
| real render, 1024×576×97 i2v, total | 130.4 s | 129.7 s | 128.6 s |

A note on methodology: `mx.get_peak_memory()` is the metric, not process RSS. MLX's Metal allocations do not appear in the process resident set on macOS — a 4 GB `mx.zeros` moves `ps rss` and `ru_maxrss` by zero on this machine — so RSS was flat at 10.70 GB (the mmap'd safetensors pages) in every configuration and says nothing useful here.

## Bit-exactness

Every claim below is `mx.array_equal`, never `allclose`.

### Commit 2 is exact by construction

A gather is a pure row permutation and the scale-shift-table add is elementwise, so the two commute exactly: `take(x, inv)[m, d] + table[i, d]` and `take(x + table, inv)[m, d]` are the same IEEE operation on the same operands. There is no reduction whose order could change.

### Commit 1 is *not* exact by construction — and that is the interesting part

The intuition "same weights, same rows, just fewer of them, so the result must be identical" is **false on MLX**. MLX dispatches different Metal GEMM kernels depending on the row count `M`, and different kernels use a different K-reduction order, so the same input row can come out ~1 bf16 ULP apart depending on how many rows were in the batch. Characterisation sweep with fixed weights and identical rows, varying only `M`:

| shape | subset result matches full result? |
|---|---|
| bf16 matmul 256 → 4096 | matches for all `M ≥ 2`; `M = 1` (gemv path) differs |
| bf16 matmul 4096 → 4096 | small-`M` kernel differs from large-`M` kernel; crossover between `M = 64` and `M = 137` |
| bf16 `nn.Linear` 4096 → 36864 | matches for `M ≥ 2`, differs at `M = 1` |
| `QuantizedLinear` q8/q4 4096 → 36864 | small-`M` path differs from the gemm path; crossover between `M = 4` and `M = 64` |

With the real modules at 20 000 tokens: a 2-row GEMM is rejected, 256 rows is accepted for the 4096-dim heads, and the 2048-dim heads need 1024 rows. There is no safe constant to hard-code — the threshold moves with output width. Anyone who attempts this optimisation will hit this, so it is worth writing down.

Two things make the shipped version exact anyway:

- **Padding.** The distinct rows are repeated up to a candidate row count (`unique → 256 → 1024 → 4096`, cheapest first) so the shrunken GEMM lands on the same kernel as the full-size one. Padding rows are pure filler — each GEMM output row is independent — so they only influence kernel selection.
- **One-time self-verification.** The first time a shape signature is seen, the code computes the *real* full-size result, checks the shrunken result against it with `mx.array_equal` (chunked, to bound peak memory), and returns **the reference** for that call. Only a proven-equal configuration is ever used afterwards. Kernel selection depends on shapes, dtypes and strides and never on values, so one verification per signature is sound. If no candidate matches exactly, the signature is permanently marked not-deduplicable and the original path runs.

Cost of that safety: one extra full-size AdaLN GEMM per shape signature per process (~0.4 s once at 20 000 tokens), then every later step is cheap. The failure mode is "no speedup", never "wrong output".

The grouping itself is also verified rather than trusted: rows are grouped by the raw bit patterns of two columns of the sinusoidal embedding, then `mx.take(reps, inverse) == flat` is checked bitwise before the grouping is used. A key collision can only ever cost the optimisation.

(Helpfully, the AdaLN projections are not quantized in the shipped checkpoints — `quantize_config.json` says `"only_transformer_blocks": true` for both q4 and q8 — so the nastier quantized-matmul thresholds never apply on this path.)

### Validation performed

**Unit** — `tests/test_adaln_dedupe.py`, 12 tests, all exact. Covers sigma patterns `uniform` / `two` (i2v) / `four` (multiple conditioning groups) / `ramp` (continuous per-token sigma, the worst case); shapes `1×8192`, `1×20000`, `1×30000`, `2×8192`; heads `4096/9` (video), `2048/9` (audio), `4096/4` and `2048/4` (AV cross-attention). Both the deduped values and the deferred per-block expansion are asserted against the pre-optimisation implementation, reproduced verbatim in the test file. Both kill switches are tested, as are the two original `_unpack_adaln` branches.

**End-to-end** — same code, same seed, same everything; the only difference between runs is the two env vars. Final video and audio latents compared by sha256 (stronger than comparing mp4 bytes):

| render | tokens (stage 1 / stage 2) | video latent sha256 | audio latent |
|---|---|---|---|
| i2v 640×480×25, q4 distilled, seed 1234 | 320 / 1 120 | `3e06b79f…` — identical across all three configurations | identical |
| i2v 1024×576×97, q4 distilled, 8+3 steps, seed 1234 | 1 872 / 7 488 | `cbc1464c…` — identical across all three configurations | identical |

("all three configurations" = both optimisations on, dedupe only, and both off.)

**Suite** — `pytest -m "not slow"` on this branch: 618 passed, 21 failed. Same 21 failures on unmodified `main` (missing local weights / a trainer import), byte-identical failure lists. `ruff check` and `ruff format --check` clean on the touched files.

## Scope and escape hatches

One new class in `adaln.py`, one branch at the top of `_unpack_adaln`, and the dedupe machinery in `model.py`. No public API change, no change to any weight key, no change to scalar-timestep behaviour, and t2v never enters this path at all. `mx.checkpoint`-based gradient checkpointing is unaffected — the carrier is captured as a constant, exactly as the eager tensor was.

| env var | default | effect |
|---|---|---|
| `LTX2_ADALN_DEDUPE=0` | on | disables the deduplicated GEMM (and therefore the deferred gather, which has nothing to defer without it) |
| `LTX2_ADALN_LAZY=0` | on | disables the deferred gather only; the deduplicated result is expanded eagerly, as in commit 1 |
| `LTX2_ADALN_DEDUPE_MIN_ROWS` | 1024 | below this many rows the GEMM is cheap and the bookkeeping is skipped |
| `LTX2_ADALN_DEDUPE_MAX_FRAC` | 0.5 | decline when the distinct rows are more than this fraction of the total |
| `LTX2_ADALN_DEDUPE_DEBUG=1` | off | trace every calibration verdict |

Degenerate inputs degrade gracefully rather than break. A genuine continuous per-token sigma ramp — which no path in the tree currently produces — still partially deduplicates for free, because per-token timesteps arrive as bfloat16 and a 20 000-token linear ramp collapses to 1 021 distinct rows; the test suite covers that case and it stays bit-identical. Anything worse than that falls out to the original code path with no measurable overhead.

Happy to split this into two PRs if you would rather review them separately, though commit 2 depends on commit 1.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1071,6 +1071,22 @@ LTX-2 mitigates both by inserting `mx.eval` at strategic points so each command 
 
 Mac Studio / M-series Ultra users who have never seen a watchdog crash may recover full lazy-graph pipelining by setting `LTX2_GEMMA_EVAL_EVERY=0` and `LTX2_DIT_EVAL_EVERY=0`. This disables all eval guards and restores maximum throughput at the cost of watchdog safety on machines where those guards were needed.
 
+### AdaLN dedupe switches (per-token timesteps only)
+
+The per-token AdaLN path deduplicates identical sigma rows into one shrunken
+GEMM (bitwise-verified per module by a one-time calibration; the calibrating
+call returns the exact reference) and defers the per-token gather into the
+blocks. Both are on by default and bit-identical by construction:
+
+- `LTX2_ADALN_DEDUPE=0` — disable the dedupe entirely.
+- `LTX2_ADALN_LAZY=0` — keep the dedupe but materialise the gather eagerly
+  (the `--low-ram` streamer does this automatically: the lazy carrier cannot
+  cross the compiled block, and the compiled block is what keeps the 48
+  per-block syncs under the Metal watchdog).
+- `LTX2_ADALN_DEDUPE_MIN_ROWS` (default 1024), `LTX2_ADALN_DEDUPE_MAX_FRAC`
+  (default 0.5) — thresholds below/above which the dedupe is not attempted.
+- `LTX2_ADALN_DEDUPE_DEBUG=1` — log calibration verdicts.
+
 ### `LTX2_GEMMA_MAX_LENGTH`
 
 Caps the padded Gemma sequence length (default `1024`). Reducing to `512` halves Gemma forward time but **shifts left-padded RoPE positions away from the LTX training distribution** — quality risk. Use only as a last resort on heavily contended systems.
--- a/packages/ltx-core-mlx/src/ltx_core_mlx/loader/block_streaming.py
+++ b/packages/ltx-core-mlx/src/ltx_core_mlx/loader/block_streaming.py
@@ -358,18 +358,7 @@ class StreamingLTXModel(nn.Module):
             # perturbations are active. This loses the compile speedup
             # but the eager block's per-step latency is dominated by
             # attention compute, so the regression is small (~5-10%).
-            #
-            # The lazy AdaLN carrier (``PerTokenAdaLNParams``) is the same
-            # class of non-tree argument: it only ever reaches the blocks
-            # when per-token timesteps are in play AND the dedupe+lazy
-            # pair is enabled, so gate on exactly that. The eager block
-            # keeps the lazy memory cut; only the compile speedup is
-            # forfeited on those steps.
-            from ltx_core_mlx.model.transformer import model as _model_mod
-
-            per_token = kwargs.get("video_timesteps") is not None or kwargs.get("audio_timesteps") is not None
-            lazy_carrier_possible = _model_mod._ADALN_DEDUPE and _model_mod._ADALN_LAZY
-            use_compiled = kwargs.get("perturbations") is None and not (per_token and lazy_carrier_possible)
+            use_compiled = kwargs.get("perturbations") is None
 
             def provider(idx: int) -> nn.Module:
                 streamer.bind(
--- a/packages/ltx-core-mlx/src/ltx_core_mlx/loader/block_streaming.py
+++ b/packages/ltx-core-mlx/src/ltx_core_mlx/loader/block_streaming.py
@@ -358,7 +358,18 @@ class StreamingLTXModel(nn.Module):
             # perturbations are active. This loses the compile speedup
             # but the eager block's per-step latency is dominated by
             # attention compute, so the regression is small (~5-10%).
-            use_compiled = kwargs.get("perturbations") is None
+            #
+            # The lazy AdaLN carrier (``PerTokenAdaLNParams``) is the same
+            # class of non-tree argument: it only ever reaches the blocks
+            # when per-token timesteps are in play AND the dedupe+lazy
+            # pair is enabled, so gate on exactly that. The eager block
+            # keeps the lazy memory cut; only the compile speedup is
+            # forfeited on those steps.
+            from ltx_core_mlx.model.transformer import model as _model_mod
+
+            per_token = kwargs.get("video_timesteps") is not None or kwargs.get("audio_timesteps") is not None
+            lazy_carrier_possible = _model_mod._ADALN_DEDUPE and _model_mod._ADALN_LAZY
+            use_compiled = kwargs.get("perturbations") is None and not (per_token and lazy_carrier_possible)
 
             def provider(idx: int) -> nn.Module:
                 streamer.bind(
--- a/packages/ltx-core-mlx/src/ltx_core_mlx/loader/block_streaming.py
+++ b/packages/ltx-core-mlx/src/ltx_core_mlx/loader/block_streaming.py
@@ -382,7 +382,22 @@ class StreamingLTXModel(nn.Module):
                 return compiled if use_compiled else shared
 
             kwargs["block_provider"] = provider
-        return self.inner(*args, **kwargs)
+        # The lazy AdaLN carrier (``PerTokenAdaLNParams``) is a non-tree
+        # argument, so it cannot cross the compiled block -- and the compiled
+        # block is load-bearing here: it is what keeps 48 per-block syncs
+        # under the Metal watchdog (see the module docstring). Rather than
+        # falling back to the eager block (forfeiting that protection), keep
+        # the GEMM dedupe but materialise its result eagerly for the duration
+        # of this forward: the lazy activation cut is forfeited under
+        # streaming, the watchdog/compile properties are not.
+        from ltx_core_mlx.model.transformer import model as _model_mod
+
+        lazy_prev = _model_mod._ADALN_LAZY
+        _model_mod._ADALN_LAZY = False
+        try:
+            return self.inner(*args, **kwargs)
+        finally:
+            _model_mod._ADALN_LAZY = lazy_prev
 
     def __getattr__(self, name: str):
         try:
--- a/packages/ltx-core-mlx/src/ltx_core_mlx/model/transformer/adaln.py
+++ b/packages/ltx-core-mlx/src/ltx_core_mlx/model/transformer/adaln.py
@@ -47,3 +47,84 @@ class AdaLayerNormSingle(nn.Module):
         embedded = self.emb(timestep_emb)
         params = self.linear(nn.silu(embedded))
         return params, embedded
+
+
+class PerTokenAdaLNParams:
+    """Per-token AdaLN parameters held in deduplicated form.
+
+    With per-token timesteps the AdaLN projection produces a
+    ``(B, N, num_params * dim)`` float32 tensor (the sinusoidal timestep
+    embedding is float32 and MLX promotes), which stays resident for the whole
+    transformer stack. At video resolutions that is gigabytes -- 2.95 GB for
+    the 9-parameter video head alone at 20 000 tokens -- even though the tensor
+    only ever holds a handful of distinct rows, one per conditioning group.
+
+    This carries the distinct rows plus the inverse index instead, and expands
+    a single ``(B, N, dim)`` parameter slice at a time, inside the block that
+    consumes it. Blocks obtain those slices through
+    ``BasicAVTransformerBlock._unpack_adaln``.
+
+    Expansion is bitwise identical to gathering up front: a gather is a pure
+    row permutation and the elementwise scale-shift-table add commutes with it
+    exactly (both operate element by element, with no reduction whose order
+    could change).
+    """
+
+    __slots__ = ("batch", "inverse", "num_params", "tokens", "unique")
+
+    def __init__(
+        self,
+        unique: mx.array,
+        inverse: mx.array,
+        batch: int,
+        tokens: int,
+        num_params: int,
+    ):
+        """
+        Args:
+            unique: (U, num_params * dim) distinct rows of the projection.
+            inverse: (B * N,) row index into ``unique`` for every token.
+            batch: B.
+            tokens: N.
+            num_params: Number of modulation parameters packed per row.
+        """
+        self.unique = unique
+        self.inverse = inverse
+        self.batch = int(batch)
+        self.tokens = int(tokens)
+        self.num_params = int(num_params)
+
+    # -- mx.array-compatible surface, so callers can introspect without
+    #    materialising anything --------------------------------------------
+    @property
+    def ndim(self) -> int:
+        return 3
+
+    @property
+    def shape(self) -> tuple[int, int, int]:
+        return (self.batch, self.tokens, int(self.unique.shape[-1]))
+
+    @property
+    def dtype(self):
+        return self.unique.dtype
+
+    @property
+    def unique_rows(self) -> int:
+        return int(self.unique.shape[0])
+
+    def gather(self) -> mx.array:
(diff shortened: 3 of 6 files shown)


dgrauet:
Thanks for the ping, and sorry this sat so long — the reason it looked
untouched is that it never actually ran. As a fork PR it was parked in
`action_required` since 3 Aug waiting for a maintainer to approve the
workflow, so there was no CI signal on it at all. I've approved it now,
and rebased it onto a green `main` first (`main` had been red since
11 Aug — hatchling 1.32 broke the editable build outright, fixed in #89,
released as 0.14.20).

Reviewed the diff in full before approving, and I want to say up front
that the design reads well. The gather/add commutation in commit 2 is
exact by construction as you describe — every output element is
`unique[inv[m], i, d] + table[i, d]` on both paths, same IEEE op, no
reduction whose order could move. The grouping is verified rather than
trusted, calibration returns the reference for the calibrating call, and
the failure mode is "no speedup" by design. I also checked that
`PerTokenAdaLNParams` can only reach `_unpack_adaln`, `_output_block`
and `compute_video_normed_sa`, all three of which handle it; the
`prompt_*` params stay plain arrays, and neither TeaCache nor
`TiledLTXModel` touches that path.

**But CI disagrees, and it disagrees on the one claim the PR rests on.**

Run: https://github.com/dgrauet/ltx-2-mlx/actions/runs/31826166400

7 of the 12 new tests fail on both Python versions:

```
tests/test_adaln_dedupe.py::test_dedupe_is_bit_identical_video
tests/test_adaln_dedupe.py::test_dedupe_is_bit_identical_other_heads
tests/test_adaln_dedupe.py::test_dedupe_repeats_are_stable_across_calls
tests/test_adaln_dedupe.py::test_kill_switch_disables_dedupe
tests/test_adaln_dedupe.py::test_lazy_unpack_is_bit_identical
tests/test_adaln_dedupe.py::test_lazy_carrier_holds_only_the_distinct_rows
tests/test_adaln_dedupe.py::test_lazy_kill_switch_materialises_eagerly
```

First failure is the very first label:

```
E   AssertionError: video/uniform/1x8192: params not bit-identical
```

The only variable that moves is the GPU:

| | you | me | CI |
|---|---|---|---|
| chip | M4 Max | **M2 Pro** | `macos-14-arm64` = **M1** |
| mlx / mlx-metal | — | 0.31.1 | 0.31.1 |
| `test_adaln_dedupe.py` | pass | **12 passed** | **7 failed** |

Same MLX on both sides of the comparison I can control, so this is not
version drift — it tracks the hardware. Which is uncomfortable, because
the calibration is exactly what is supposed to make hardware differences
a non-event: a kernel pair that disagrees should be *rejected* and fall
back, never accepted and used.

Two observations that may save you time, offered as leads rather than
conclusions — it's your code and you have the M4 Max bench, so I'd
rather hand this back early than spend hours in a cache whose intent I'm
guessing at:

1. **`test_kill_switch_disables_dedupe` is in the failing set.** With the
   dedupe disabled the result should be trivially identical, so at least
   part of this is not kernel selection at all. That smells like state
   leaking between tests through the module-global `_ADALN_DEDUPE_PLAN`.

2. **The plan is keyed on shapes, not on weights.** `_adaln_signature`
   carries `(M, D, padded_rows, num_params, weight.shape, weight.dtype,
   linear type, flat.dtype)` — nothing identifying *which* module. The
   header argues one verification per signature is sound because kernel
   selection never depends on values. That's true of *which kernel runs*,
   but what the check actually establishes is *bit-equality between two
   kernels*, and that is value-dependent: two reduction orders can agree
   on the calibration input and diverge on a later one. In `_module(dim,
   num_params, seed)` the tests build fresh random weights per test while
   the plan persists across them, so an ACCEPTED verdict earned under one
   weight set gets reused under another. On M4 Max the kernels may
   coincide often enough to hide it; on M1 they apparently don't.

   The stronger justification for the one-shot check, if you want to keep
   it, is probably not "kernel selection is value-independent" but
   "agreement across M x 36864 float32 values is overwhelming evidence
   that the *same* kernel was selected, in which case equality is
   universal". That reasoning holds — but only if the verdict cannot be
   transported to a different GEMM than the one it was established on.

Also worth flagging: for the `uniform` pattern `u == 1`, so the first
candidate from `_candidate_rows` is a 1-row GEMM — the gemv path your
own characterisation table lists as differing from the gemm path. That's
the case that fails first.

For context on why this matters here specifically: this repo already has
a case of exactly this shape in #40 — q4 distilled producing a
mosaic/rainbow grid on M2/M3 while clean on M4 Max, identical MLX
0.31.1. Numerical behaviour that validates on M4 Max and breaks on older
silicon is a pattern we've paid for once already, which is why I'd like
the guarantee to be robust rather than empirically-true-on-my-machine.

Happy to keep the CI approvals coming so you get a signal on each push —
just ping me. Nothing here changes my read that the underlying idea is
worth landing; a 63% cut in step activations with no wall-clock cost is
well worth the care it's taking to get right.

dgrauet:
Took this over as offered — the underlying idea was worth landing, and it now does. What changed on top of your two commits (kept verbatim, rebased onto current main + mlx 0.32.2):

1. **The M1-CI failure root cause**: the calibration verdict lived in a module-global dict keyed on shapes only, so an ACCEPTED verdict earned under one module's weights was transported to any same-shaped module with different weights — bit-equality evidence doesn't transport. The plan now lives on each `AdaLayerNormSingle` instance (`__dict__`, bypassing `Module.__setattr__` so it never reaches `parameters()`), and the signature also keys on the weight array's identity (a fine-tune step invalidates stale verdicts). Two of your lazy tests unknowingly depended on the transport (carrier expected on a fresh module's first call) — they now warm the calibrating call explicitly. New regression tests pin non-transport and the `parameters()` non-leak, mutation-verified.
2. **A latent `--low-ram` crash**: the lazy carrier is a non-tree argument, so conditioned generations under streaming died in `mx.compile`. The streamer now disables the lazy gather (keeping your GEMM dedupe) for streamed forwards — the compiled block is what keeps 48 per-block syncs under the Metal watchdog, so it must not be forfeited.
3. Docs for the `LTX2_ADALN_*` switches, empirical wording on the calibration argument.

Empirical validation on the 2.5 q8 pack (M2 Pro, mlx 0.32.2, 512×512×25 I2V, seed 81647281): dedupe ON == OFF **bit-for-bit in both block flavours** (eager `83a051c9…`, streamed/compiled `31b239a7…`); the uniform-T2V SHA-gate is untouched; peak RSS 12.7 vs 13.1 GB at this small config. Suites 816 + slow green.

Thanks for the contribution — merging on green CI. Your commits stay yours in the history.

dgrauet:
@mrbizarro — one follow-up now that this is merged, if you're willing: a counter-bench on your M4 Max to settle the *default*.

On M2 Pro at production res the trade currently reads against the default-on (all three arms bit-identical, \`e0c3567c…\`):

| Arm | Time | Peak RSS |
|---|---|---|
| plain (\`LTX2_ADALN_DEDUPE=0\`) | **312.8 s** | 12.76 GB |
| dedupe+lazy (default) | 339.5 s (+8.5 %) | **12.27 GB** (−0.49) |
| dedupe w/o lazy (\`LTX2_ADALN_LAZY=0\`) | 420.0 s | 13.11 GB |

Your multi-GB/M4 numbers predate the take-over changes, so a fresh run with the same protocol would let us decide default-on vs opt-in on evidence from both hardware classes. Protocol (current \`main\`, mlx 0.32.2):

```bash
# arm 1 — default (dedupe+lazy)
/usr/bin/time -l ltx-2-mlx generate --model <2.5-q8-pack> --distilled \
  --seed 81647281 -H 480 -W 704 -f 97 --frame-rate 24 \
  --image <any-anchor.png> -p "a heavy wooden door creaks slowly open in an old stone house, dust swirling in a shaft of afternoon light" \
  -o on.mp4
# arm 2 — plain
LTX2_ADALN_DEDUPE=0 /usr/bin/time -l ltx-2-mlx generate ... -o off.mp4
# arm 3 — dedupe only
LTX2_ADALN_LAZY=0 /usr/bin/time -l ltx-2-mlx generate ... -o dedupe_only.mp4
```

Report \`Time:\`, \`maximum resident set size\`, and the three SHA-256 (they must match each other on your machine too — I2V is what exercises the per-token path; no \`--low-ram\`, streaming now runs the plain path by design, see #119). If your numbers show the win your original bench did, default-on stays; if they look like mine, we'll flip to opt-in. Either way the mechanism remains available and bit-exact.

PR #88, opened 2026-08-11
Feat: stepwise previews during generation

plz12345:
## Summary

A generation is opaque until the final VAE decode, which for a long clip is many minutes away. This adds opt-in stepwise previews: every N denoising steps, decode a short **contiguous window** of latent frames from the in-progress x0 prediction and write it as a self-contained animated WebP — a couple of seconds of real motion at the current denoise quality, openable mid-run.

```
--stepwise-image-output-dir DIR   Write per-step preview clips here (off by default)
--stepwise-interval N             Preview every N steps (default: 1; last step always previewed)
--stepwise-frames N               Latent frames per preview (default: 8 -> 57 pixel frames, ~2.3s)
--stepwise-frame I                Latent frame the window is centred on (default: middle)
```

```bash
ltx-2-mlx generate -p "a cat walking" -o out.mp4 --frame-rate 24 \
  --stepwise-image-output-dir ./previews
```

```
previews/
  seed_42_s1_step001of030.webp   # 57 frames of motion, this step
  seed_42_s1_step002of030.webp
  ...
  seed_42_s2_step003of003.webp
```

Available on every generating subcommand: `generate`, `a2v`, `retake`, `extend`, `keyframe`, `ic-lora`, `hdr-ic-lora`, `lipdub`. `--stepwise-image-output-dir` matches mflux's flag name deliberately, for tooling that already drives both.

## Why a window rather than a single frame

A single decoded frame cannot show temporal coherence, flicker, or whether a subject moves sensibly — which is most of what goes wrong in a video model. It is also the worst point on the cost curve: the VAE upsamples time 8×, so `N` latent frames decode to `8N-7` pixel frames, and one frame pays for the whole up-block stack while throwing the temporal upsampling away.

Measured on a 704×448 clip:

| latent frames | pixel frames | motion @25fps | stage 1 | stage 2 |
|---|---|---|---|---|
| 1 | 1 | — | 0.06s | 0.21s |
| 3 | 17 | 0.68s | 0.49s | 1.90s |
| **8** (default) | **57** | **2.3s** | 1.55s | 6.13s |
| 16 | 121 | 4.8s | 3.23s | 13.0s |

8 latent frames cost ~26× one frame but yield 57 of them. **Cost is independent of clip length** — decoding 8 of 60 latent frames costs the same as 8 of 16 — so previews get proportionally cheaper the longer the clip: roughly 7% of a 15-minute two-stage run, but closer to 50% of a short distilled one, where `--stepwise-interval` is the lever.

## Design notes

**One file per step, written once, never rewritten.** The obvious design — append each step to one growing animation — is quadratic in the step count, and the rewrite cost lands on exactly the long runs previews are for. Callers that want the whole progression play the chunks back to back: the same motion span replayed once per step, sharpening as it converges.

**A new `on_step`, not a widened `tap`.** The three samplers take a new optional callable rather than reusing `tap`, which is teacache calibration plumbing: `tap` carries block residuals instead of latents, fires *before* guidance combination and mask blending, has a live consumer in `calibrate_teacache.py`, forces gate-signal computation and residual capture merely by being set, and does not exist on `denoise_loop` at all. `on_step` receives the guided, mask-blended x0 — what the step actually produced.

**The samplers stay ignorant of pipelines.** `StepwisePreview.bind()` closes over this loop's latent geometry, decoder and patchifier and hands back a plain `on_step(step_idx, num_steps, video_x0, sigma)`. Multi-stage pipelines bind once per stage — they run at different resolutions — and tag files `_s1` / `_s2`.

**No `generate_and_save` signature was touched.** The handler hangs off `BasePipeline` and the CLI sets it as an attribute alongside `verbose`. This matters because `ti2vid_one_stage` swallows unknown kwargs while `ti2vid_two_stages` would raise on them; a kwarg would have needed threading through every pipeline's signature to be safe.

**Failure is contained.** A preview exception disables the handler for the rest of the run — it can neither abort a long generation nor slow every remaining step. Chunks are written to a temp file and `os.replace`d into place, so a reader polling the directory can never open a half-written file. The output directory is created and write-probed at CLI parse time, so a bad path fails in the first second rather than minutes in.

## Caveats (documented in README)

- **Memory**: previews keep the VAE decoder resident through denoising, which the pipelines otherwise deliberately avoid. Combining with `--low-ram` works but is self-defeating, and prints a warning.
- The window is centred on the middle of the clip by default, because latent frame 0 is the clean conditioning image on I2V runs and never changes.
- Decoding a window in isolation uses different boundary padding than the full-volume decode, so the outermost frames are approximate. Fine for judging motion, not final quality.
- Flags are marked `[EXPERIMENTAL]` in `--help`; the file naming and defaults may change.

## Scope / safety

- Entirely opt-in: with the flag absent, `pipe.stepwise is None`, `_stepwise_hook` returns `None`, and each sampler's `on_step is not None` check is the only added work in the loop.
- Verified **byte-identical** MP4 output with and without the flag, on both the single-frame and windowed designs.
- End-to-end run: 11 chunks × 57 frames on a 5s distilled generation.
- 36 new tests in `tests/test_stepwise.py` covering window resolution and clamping, pixel conversion, per-stage binding, chunk naming and atomic writes, interval behaviour (including the always-previewed last step), failure isolation, and the sampler hook wiring.
- Gates pass locally: `ruff check packages/ tests/`, `ruff format --check packages/ tests/`, full suite `643 passed, 22 skipped`.


dgrauet:
Reviewed, and this one is in good shape — I'd merge it with one change,
detailed at the end.

First, housekeeping: I refreshed the branch onto current `main`. Your CI
was green on 11 Aug but that was against a base that went red later the
same day (hatchling 1.32 broke the editable build outright — fixed in
#89, released as 0.14.20). It's green again on the current base, so the
signal is real now. Your `chore: format python blocks in CLAUDE.md for
ruff 0.16` commit is subsumed by #89; thanks for the diagnosis on #85,
it was exactly right and it unblocked every open PR at once.

## Verified

| | |
|---|---|
| CI on refreshed base (`macos-14-arm64`) | 5/5 pass |
| `tests/test_stepwise.py` locally (M2 Pro) | 36 passed, 2.2s |
| Full suite | 661 passed (625 baseline + your 36), 1 skipped |
| Pre-existing failures | same 3 ffmpeg `communicate()` ones, unchanged |
| File overlap with #83 | none |

## What reads well

The sampler hooks are three lines each and sit after the mask blend and
before the Euler step, so `on_step` receives what the step actually
produced rather than a pre-guidance intermediate. Inert by construction:
`pipe.stepwise` is `None` by default, `_build_stepwise` returns `None`
without the flag, and the only added work in the loop is an `is not None`
test.

Declining to widen `tap` was the right call, and for the reasons you
give — it carries block residuals, fires before guidance combination,
has a live consumer in `calibrate_teacache.py`, and forces residual
capture merely by being set. Keeping `on_step` separate keeps both
honest.

Three things I checked that could plausibly have bitten, all fine:

- `decoder_block.load()` runs on every preview, but `load()` memoises
  (`if self._decoder is not None: return self._decoder`), so it's one
  load, not one per step.
- `args.seed` is resolved from `-1` at cli.py:713, before
  `_build_stepwise` is called, so filenames carry the real seed rather
  than `seed_-1_`.
- `_decode_window` truncates to `F*H*W` before `unpatchify`, which
  correctly handles multi-anchor and reference-latent conditioning where
  tokens are appended past the grid.

## The one change I'd ask for

**Emit the memory warning unconditionally, not only under `--low-ram`.**

Previews keep the VAE decoder resident through denoising, so the DiT and
the decoder are live at the same time. That is precisely the shape of
#74 — *"Pipelines keep the DiT loaded across VAE decode — silent SIGKILL
(jetsam) on 32 GB Macs"* — which we only fixed three weeks ago in
0.14.19.

What makes this worth a line of code is that your failure containment
cannot cover this particular failure. `_disable()` catches `Exception`,
but a jetsam kill is not an exception — the process dies with no
traceback and no Python-level signal at all. So on a 32 GB Mac *without*
`--low-ram`, a user who enables previews on a long clip gets a silent
death and nothing pointing at previews as the cause. The `--help` text
does disclose the cost, so this isn't an omission; it's that the one
case where the user most needs the hint is the case where they get no
output at all.

Moving the existing `print` out of the `if getattr(args, "low_ram", ...)`
branch (and keeping the sharper `--low-ram` wording as an extra line)
would cover it.

## Two nits, take or leave

- `on_step: callable | None` uses the lowercase builtin rather than
  `Callable`. Technically wrong as an annotation — but it matches the
  existing `tap: callable | None` two functions up, so it's local
  convention rather than anything you introduced. Only worth touching if
  you want to fix both.
- No file overlap with #83, though it optimises a neighbouring frame
  conversion (`memoryview` into ffmpeg, versus `memoryview` into PIL
  here). They can land in either order; just flagging in case you'd
  rather converge on one helper later.

Happy to merge as soon as the warning moves. Nice piece of work — the
reasoning in the PR description about why a window beats a single frame,
and why one-file-per-step beats a growing animation, is the kind of thing
that makes a feature reviewable in one pass.

plz12345:
Both done in `111f68e`.

**Warning is unconditional.** You're right that the containment can't reach this one — `_disable()` catches exceptions and jetsam isn't one. The `print` moved out of the `--low-ram` branch, with the sharper wording kept as a second line:

```
[stepwise] warning: previews keep the VAE decoder resident through denoising, so it and the transformer are in memory at once. On a memory-tight machine this can get the process killed outright, with no error.
[stepwise] warning: raising peak memory is the opposite of what --low-ram is for; consider dropping one of the two.
```

Second line only under `--low-ram`. Your reasoning is a comment above the print so nobody tidies it back into the branch. README updated; three tests pin it (warns without `--low-ram`, second line with it, silent when previews are off).

**`callable`.** Fixed both, via named aliases in `samplers.py` so the contract has one definition:

```python
OnStepFn = Callable[[int, int, mx.array, float], None]
TapFn = Callable[[int, mx.array | None, mx.array, mx.array], None]
```

Reused in `a2vid_two_stage`, `_base._stepwise_hook` (was bare `Callable`) and `stepwise.bind` (was the expanded form). I left the ~11 other lowercase ones in `ltx-core-mlx/model.py` and the pipelines' `tap` args alone — a separate cleanup rather than pulling another package into this diff. Say the word if you'd rather have them here.

On #83: agreed, either order. Happy to converge on one frame-conversion helper once both land.

Gates: `ruff check` + `ruff format --check` clean, full suite 646 passed / 22 skipped, `test_stepwise.py` 39 (was 36).

PR #128, opened 2026-09-08
fix: stage-2 sigma schedule must always terminate at 0.0

jsirish:
## What

`STAGE_2_SIGMAS` is a fixed 4-value table (3 steps) whose terminal value is `0.0`. Every
two-stage pipeline prefix-slices it as `STAGE_2_SIGMAS[: stage2_steps + 1]`, which **drops
that terminal `0.0`** whenever `stage2_steps < 3`:

```python
>>> STAGE_2_SIGMAS[: 2 + 1]
[0.909375, 0.725, 0.421875]   # never reaches 0.0
```

`denoise_loop` therefore stops at sigma 0.42 and hands the VAE a latent that is still
~42% noised. The decoder renders that as heavy colour speckle across the whole frame.

This adds `scheduler.stage2_sigmas()`, which always terminates at `0.0` for any step count,
and routes the seven affected two-stage call sites through it.

## Why it matters

`--stage2-steps` is a documented flag, so any value below 3 silently produces corrupted
output rather than a faster-but-rougher render. Found downstream while chasing exactly that:
an a2v "fast" preset at `stage2_steps=2` produced speckle, which looked like a TeaCache
regression until the sigma schedule turned out to be the cause.

## Behaviour

| `stage2_steps` | before | after |
|---|---|---|
| `None` / `0` | full 3-step table | **unchanged** |
| `1` | `[0.909, 0.725]` — no terminal 0 | `[0.909, 0.0]` |
| `2` | `[0.909, 0.725, 0.422]` — no terminal 0 | `[0.909, 0.725, 0.0]` |
| `3` | full table | **unchanged (byte-identical)** |
| `>= 4` | full table | clamped to the table's 3 steps (no phantom zero-length final step) |

The default path is byte-identical, so nothing changes for anyone not passing
`--stage2-steps 1` or `2`.

## Deliberately NOT changed: the LTX-2.5 path

`DistilledPipeline` selects `LTX_2_5_STAGE_2_DISTILLED_SIGMAS` on 2.5 packs, and that table
carries the same prefix-slice. It **looks** like the same bug, and my first attempt fixed it
too — but that breaks two of your tests on purpose:

- `test_stage_step_truncation_applies_to_the_25_tables` asserts the truncated slice directly
- `test_25_pack_stage2_stays_deterministic` asserts *object identity* of the table, with the
  comment "identity, not equality: the 2.3 table is value-identical"

Both read as intentional, and I have only ever reproduced the speckle on the dev two-stage
path — never on a 2.5 pack. So the 2.5 branch keeps its existing behaviour here and the
scoping is documented in `stage2_sigmas`'s docstring. Happy to extend it if you confirm the
2.5 path should terminate at 0.0 as well; that felt like your call rather than mine.

## Tests

New `tests/test_scheduler.py` cases cover every step count including the degenerate ones.
Full suite: **813 passed, 58 skipped**, `ruff check` clean.

### DIFF
--- a/packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/a2vid_two_stage.py
+++ b/packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/a2vid_two_stage.py
@@ -31,7 +31,7 @@ from ltx_core_mlx.model.transformer.model import X0Model
 from ltx_core_mlx.utils.audio import load_audio
 from ltx_core_mlx.utils.memory import aggressive_cleanup
 from ltx_core_mlx.utils.positions import compute_audio_positions, compute_audio_token_count, compute_video_positions
-from ltx_pipelines_mlx.scheduler import STAGE_2_SIGMAS, ltx2_schedule
+from ltx_pipelines_mlx.scheduler import ltx2_schedule, stage2_sigmas
 from ltx_pipelines_mlx.ti2vid_two_stages import DEFAULT_CFG_SCALE, TI2VidTwoStagesPipeline
 from ltx_pipelines_mlx.utils.helpers import create_noised_state
 from ltx_pipelines_mlx.utils.samplers import OnStepFn, denoise_loop, guided_denoise_loop
@@ -312,7 +312,7 @@ class A2VidPipelineTwoStage(TI2VidTwoStagesPipeline):
         # --- Stage 2: Refine at full resolution (no CFG) ---
         video_tokens, _ = self.video_patchifier.patchify(video_upscaled)
 
-        sigmas_2 = STAGE_2_SIGMAS[: stage2_steps + 1] if stage2_steps else STAGE_2_SIGMAS
+        sigmas_2 = stage2_sigmas(stage2_steps)
         start_sigma = sigmas_2[0]
 
         video_positions_2 = compute_video_positions(F, H_full, W_full)
--- a/packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/distilled.py
+++ b/packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/distilled.py
@@ -40,7 +40,7 @@ from .scheduler import (
     DISTILLED_SIGMAS,
     LTX_2_5_DISTILLED_SIGMAS,
     LTX_2_5_STAGE_2_DISTILLED_SIGMAS,
-    STAGE_2_SIGMAS,
+    stage2_sigmas,
 )
 from .ti2vid_two_stages import TI2VidTwoStagesPipeline
 from .utils.helpers import create_noised_state
@@ -392,8 +392,14 @@ class DistilledPipeline(TI2VidTwoStagesPipeline):
 
         # --- Stage 2: full resolution refine (no LoRA swap — already distilled) ---
         video_tokens, _ = self.video_patchifier.patchify(video_upscaled)
-        stage2_table = LTX_2_5_STAGE_2_DISTILLED_SIGMAS if self._is_25 else STAGE_2_SIGMAS
-        sigmas_2 = stage2_table[: stage2_steps + 1] if stage2_steps else stage2_table
+        if self._is_25:
+            # Upstream's 2.5 path keeps its own slice: its tests assert both the
+            # truncation and the table's object identity, and #33's speckle was
+            # never reproduced on a 2.5 pack. See scheduler.stage2_sigmas.
+            stage2_table = LTX_2_5_STAGE_2_DISTILLED_SIGMAS
+            sigmas_2 = stage2_table[: stage2_steps + 1] if stage2_steps else stage2_table
+        else:
+            sigmas_2 = stage2_sigmas(stage2_steps)
         # Upstream renoises the upscaled stage-1 latent at ``stage_2_sigmas[0]``
         # (``ModalitySpec(noise_scale=stage_2_sigmas[0].item())``); our
         # ``create_noised_state(sigma=...)`` below is that same mechanism.
--- a/packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/ic_lora.py
+++ b/packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/ic_lora.py
@@ -41,7 +41,7 @@ from ltx_pipelines_mlx.iclora_utils import (
     append_ic_lora_reference_video_conditionings,
     read_lora_reference_downscale_factor,
 )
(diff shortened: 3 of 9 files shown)


dgrauet (changes requested):
Thanks for this — the bug is real and well diagnosed. `STAGE_2_SIGMAS[: stage2_steps + 1]` with `stage2_steps=2` yields `[0.909, 0.725, 0.422]`, so the VAE gets a latent that is still ~42% noised. The default path (3 steps / `None`) stays byte-identical, and the tests pin that. Nothing to change on the mechanics of the 2.3 path.

Two things I'd like addressed before merging, plus a couple of nits.

### 1. The 2.5 path should be fixed as well

The two tests you cite as "intentional" aren't pinning a no-terminal-zero schedule:

- `test_stage_step_truncation_applies_to_the_25_tables` was written in #106 to pin the **routing** to the 2.5 table (i.e. that `stage2_steps` is honoured on 2.5 packs at all), not to consecrate the truncation semantics. It just copied the slice that existed at the time.
- `test_25_pack_stage2_stays_deterministic` only asserts object identity on the **untruncated** path.

Both tables are value-identical, so `--stage2-steps 2` on a 2.5 pack produces exactly the same speckle. Suggested shape:

```python
def stage2_sigmas(stage2_steps: int | None = None, table: list[float] = STAGE_2_SIGMAS) -> list[float]:
    if not stage2_steps:
        return table  # same object: keeps the identity assertion in test_ltx25_distilled
    return table[: min(stage2_steps, len(table) - 1)] + [0.0]
```

Then route the `_is_25` branch in `distilled.py` through it with `LTX_2_5_STAGE_2_DISTILLED_SIGMAS`, and update the truncation assertion to `[0.909375, 0.725, 0.0]`. That also removes the `if head and head[-1] == 0.0` special case.

### 2. Two references in the text are wrong

- The docstring says "upstream's tests assert that slice AND the table's object identity". Those are this repo's tests, not Lightricks'. Upstream has no `stage2_steps` equivalent at all — `STAGE_2_DISTILLED_SIGMAS` is a fixed table there.
- The comment in `distilled.py` mentions "#33's speckle". #33 in this repo is the 0.14.9 release PR; I assume that's an issue number from your downstream tracker. Please drop it.

### Nits

- Since upstream has no short stage-2 schedule (its only 2-step one is the TDP multi-GPU runner's `[0.625, 0.4, 0.0]`, which is *not* "table head + 0"), the `head + [0.0]` choice is reasonable but not a calibrated schedule. A sentence in the docstring saying that `stage2_steps < 3` is a fast, uncalibrated mode would set expectations.
- Out of scope for this PR, but same bug class for a follow-up: the stage-1 truncation `DISTILLED_SIGMAS[: stage1_steps + 1]` in `ic_lora.py`, `keyframe_interpolation.py`, `lipdub.py` and `validation_sampler.py` also never reaches 0 (less visible because stage 2 renoises at 0.909, but the upscaled latent is still partially noised), and `get_sigma_schedule` slices `[:num_steps]`, returning N sigmas for N requested steps.

dgrauet:
Follow-up after a second, deeper pass. These refine the review above; the two blocking points stand.

### Prefer a tail-slice over `head + [0.0]`

The repo already has a truncation convention for short distilled schedules: `--refine-steps N` in `ic_lora.py:632` takes the **tail** of the distilled ladder (`DISTILLED_SIGMAS[-(n+1):]`), documented in CLAUDE.md as "N maps onto the distilled sigma tail". Since `STAGE_2_SIGMAS == DISTILLED_SIGMAS[-4:]` (asserted in `test_ltx25_ancestral.py`), `table[-(n+1):]` gives:

| n | head + 0 (this PR) | tail-slice |
|---|---|---|
| 1 | `[0.909, 0.0]` | `[0.422, 0.0]` |
| 2 | `[0.909, 0.725, 0.0]` | `[0.725, 0.422, 0.0]` |
| 3 | full table | full table (byte-identical) |

The tail-slice only ever uses sigma pairs that exist in the calibrated ladder, whereas `0.725 → 0` and `0.909 → 0` exist nowhere in it. It also reconciles `--stage2-steps` with `--refine-steps` (right now the two knobs in the same `if/else` in `ic_lora.py` would produce different schedules for equal N), and it removes the `head[-1] == 0.0` special case. `test_two_steps_keeps_calibrated_head` currently locks the head choice in without any render showing it is "rougher" rather than differently wrong, so please switch that test along with the helper.

### Same bug class, wider than stage 2

- `ic-lora --single-stage --stage1-steps 4` → `DISTILLED_SIGMAS[:5]` ends at 0.975 and `ic_lora.py:527-531` returns that latent **straight to the decoder** with no stage-2 renoise to mask it. Near-pure-noise output. Same at `lipdub.py:243`, `keyframe_interpolation.py:268`, `distilled.py:331`.
- `validation_sampler.py:186` in the trainer: `num_inference_steps: 4` in a YAML makes every validation video decode as speckle, which reads as "my LoRA is diverging".
- `get_sigma_schedule("stage_2", num_steps)` still does `sigmas[:num_steps]` (no +1, no terminal 0) and is a public export.

Given that, a single table-generic helper (e.g. `truncate_sigmas(table, steps)`) routed through every stage-1 and stage-2 site, 2.3 and 2.5, fixes the class. A helper hard-bound to `STAGE_2_SIGMAS` only fixes the sites you grepped for. I'm fine with the stage-1 sites landing in a follow-up PR if you'd rather keep this one small, but the helper should be table-generic from the start.

### Smaller items

- Negative `stage2_steps` is not rejected and the CLI has no range check: `stage2_sigmas(-4)` → `[0.0]`, so stage 2 becomes a silent no-op decoding the raw upsampler output. A `if stage2_steps < 0: raise ValueError` closes it.
- Docstring `Returns:` says "length num_steps + 1"; it is clamped to 4 for `steps >= 4`, as your own clamp test asserts. Also worth a warning on `steps > 3` — stepwise previews will be named `_stepNNNof003` with no hint why.
- The new tests only cover the pure helper. Nothing asserts a pipeline routes `stage2_steps < 3` through it, so re-inlining the old slice at any of the seven call sites keeps CI green. Extending `_make_stubbed_pipeline` in `tests/test_ltx25_distilled.py` with `stage2_steps=2` on the 2.3 path and asserting `euler.calls[1]["sigmas"][-1] == 0.0` would guard the actual fix.

dgrauet:
Closing in favour of #141, which fixes the same terminal-sigma bug across both the 2.3 and 2.5 tables and both stages (the two points requested here on 2026-09-12) and carries measurements. Thanks for the original diagnosis — it was right; the follow-up landed there.

PR #129, opened 2026-09-08
feat: wire TeaCache into A2VidPipelineTwoStage stage-1

jsirish:
## What

`--enable-teacache` is wired onto every `generate` subcommand by `_add_generation_args`, and
`a2v` accepts the flag — but `A2VidPipelineTwoStage` never built a controller or passed one
down, so on `a2v` the flag was **accepted and silently ignored**.

This threads it through:

- `_denoise_stage1` takes a `teacache_controller` param and forwards it as
  `teacache=` to `guided_denoise_loop`
- `generate_and_save` takes `enable_teacache` / `teacache_thresh` and builds the controller
  via the existing `_build_teacache_controller` from `ti2vid_two_stages`
- `cli.py::_cmd_a2v` extracts both from args and passes them on (the flag itself was already
  wired; only the handler was missing)

No new machinery — `TI2VidTwoStagesPipeline` already carries the calibrated Euler
coefficients (`LTX2_TEACACHE_COEFFICIENTS`) and the builder, and `guided_denoise_loop`
already takes `teacache=`. This just connects the a2v pipeline to them.

## Note on calibration

The coefficients are the ones calibrated for `guided_denoise_loop` (Euler), which is the loop
a2v stage 1 actually uses — so they apply directly. Per CLAUDE.md, calibration is
scheduler-specific, and a2v stage 1 is Euler, not res_2s. Off by default (`enable_teacache=False`),
so this is strictly opt-in.

## Measured effect (a2v, 512×320, 113 frames, M3 Ultra, q8)

Same seed, `enable_teacache` the only variable:

| stage-1 steps | off | on | speedup |
|---|---|---|---|
| 18 | 162.5s | 158.9s | 1.02× — **noise** |
| 30 | 234.2s | 180.8s | **1.30×** (23% saved) |

Worth stating plainly: **at low step counts this buys nothing.** The gain only appears once
there are enough steps for cached skips to amortise past warmup, which matches the 30-step
regime the existing coefficients were calibrated in. It is also below the 1.46× documented for
`--two-stage`, which is what I would expect — a2v stage 1 carries the audio branch and its
extra cross-attention through every non-skipped step.

Outputs at the same seed are **not** byte-identical between off and on, which is the check that
distinguishes "the controller is actually reaching the denoise loop" from "the kwarg was
accepted and silently dropped".

## Tests

`A2VidPipelineTwoStage` had **no test coverage at all**, so this adds
`tests/test_a2vid_teacache.py` — three cases pinning the wiring at `_denoise_stage1` with a
spy on `guided_denoise_loop`:

- the controller reaches the loop as `teacache=` (asserted by identity, since the controller
  carries calibrated per-model coefficients)
- omitting it yields `teacache=None` rather than a default controller
- the `on_step` hook still threads through — it shares these call sites, and neither may
  displace the other

Assertions sit at `_denoise_stage1` rather than through `generate_and_save` deliberately:
the behaviour under test is one keyword argument's journey into the denoise loop, and
stubbing the full pipeline (VAE, text encoder, audio encoder, upsampler) to observe it would
mostly test the stubs.

Verified the three fail against `main`'s `a2vid_two_stage.py` and pass with this change.
Full suite: **800 passed, 58 skipped**, `ruff check` clean.

### DIFF
--- a/packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/a2vid_two_stage.py
+++ b/packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/a2vid_two_stage.py
@@ -68,6 +68,7 @@ class A2VidPipelineTwoStage(TI2VidTwoStagesPipeline):
         cfg_scale: float = 3.0,
         stg_scale: float = 1.0,
         on_step: OnStepFn | None = None,
+        teacache_controller=None,
     ) -> object:
         """Run Stage 1 denoising with Euler + CFG. Override for HQ (res2s)."""
         # Video: full guidance (ref LTX_2_3_PARAMS)
@@ -95,6 +96,7 @@ class A2VidPipelineTwoStage(TI2VidTwoStagesPipeline):
             audio_guider_factory=audio_factory,
             sigmas=sigmas,
             on_step=on_step,
+            teacache=teacache_controller,
         )
 
     def generate_and_save(
@@ -116,6 +118,8 @@ class A2VidPipelineTwoStage(TI2VidTwoStagesPipeline):
         images=None,
         audio_start_time: float = 0.0,
         audio_max_duration: float | None = None,
+        enable_teacache: bool = False,
+        teacache_thresh: float | None = None,
     ) -> str:
         """Generate video from audio and save to file.
 
@@ -252,6 +256,12 @@ class A2VidPipelineTwoStage(TI2VidTwoStagesPipeline):
         sigmas_1 = ltx2_schedule(stage1_steps, num_tokens=num_tokens)
         x0_model = X0Model(self.dit)
 
+        teacache_controller = None
+        if enable_teacache:
+            from ltx_pipelines_mlx.ti2vid_two_stages import _build_teacache_controller
+            teacache_controller = _build_teacache_controller(stage1_steps, teacache_thresh)
+            teacache_controller.reset()
+
         output_1 = self._denoise_stage1(
             x0_model=x0_model,
             video_state=video_state_1,
@@ -264,6 +274,7 @@ class A2VidPipelineTwoStage(TI2VidTwoStagesPipeline):
             cfg_scale=cfg_scale,
             stg_scale=stg_scale,
             on_step=self._stepwise_hook(F, H_half, W_half, stage=1),
+            teacache_controller=teacache_controller,
         )
         if self.low_memory:
             aggressive_cleanup()
--- a/packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/cli.py
+++ b/packages/ltx-pipelines-mlx/src/ltx_pipelines_mlx/cli.py
@@ -1092,6 +1092,10 @@ def _cmd_a2v(args: argparse.Namespace) -> None:
         kwargs["cfg_scale"] = args.cfg_scale
     if args.stg_scale is not None:
         kwargs["stg_scale"] = args.stg_scale
+    if getattr(args, "enable_teacache", False):
+        kwargs["enable_teacache"] = True
+        if getattr(args, "teacache_thresh", None) is not None:
+            kwargs["teacache_thresh"] = args.teacache_thresh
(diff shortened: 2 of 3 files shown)


dgrauet (changes requested):
Thanks — the wiring is right, the measured 1.30× at 30 steps is a real e2e check, and the "not byte-identical" test is exactly the signal I'd want for "the controller actually reaches the loop". One blocking gap, then small things.

### Blocking: the LTX-2.5 guard is missing

`TI2VidTwoStagesPipeline.generate_two_stage` and the HQ pipeline both call `self._check_teacache_supported(enable_teacache)` (`_base.py:234`) before building the controller, because the Euler coefficients are calibrated on 2.3 and the 2.5 sampler dynamics differ. `A2VidPipelineTwoStage.generate_and_save` doesn't, so on a 2.5 pack `a2v --enable-teacache` silently runs uncalibrated skips instead of raising. Please add the call right after the `audio_path` check, and a test for it — `tests/test_ltx25_distilled.py::test_teacache_rejected_on_25_pack` is the shape to copy.

### Small things

- `teacache_controller=None` needs a type hint (`TeaCacheController | None`); type hints are mandatory across the repo.
- `generate_and_save`'s docstring doesn't list the two new args. The wording in `ti2vid_two_stages.py:398-404` can be copied.
- The function-local `from ltx_pipelines_mlx.ti2vid_two_stages import _build_teacache_controller` can be a top-level import — the module already imports `DEFAULT_CFG_SCALE` and `TI2VidTwoStagesPipeline` from there, so there is no cycle to dodge.
- `cli.py`: `getattr(args, "enable_teacache", False)` is defensive for a flag `_add_generation_args` always adds; the rest of `_cmd_a2v` uses `args.x` directly. Match that.
- The `--enable-teacache` help string still says "Two-stage (--two-stage / --two-stages-hq) only". Update it, and add `a2v` to the TeaCache coverage note in CLAUDE.md / `docs/PIPELINES.md`.

### One thing to state in the PR

a2v stage 1 runs with `stg_scale=1.0` by default, whereas the coefficients were calibrated on `--two-stage` CFG-only (T2V, `stg_scale=0.0`). The gate signal is block 0's modulated input on the cond pass, so STG doesn't change the decision, and `guided_denoise_loop` caches the `ptb` pass alongside `cond`/`uncond` — it works, and your 1.30× confirms it. But the skip-rate regime is different from the documented one, and a2v is judged on audio/visual sync. A one-line qualitative note on the 30-step on/off pair (sync preserved? any drift?) would be enough for me; no need for a PSNR run.

dgrauet:
Two additions after a second pass, both about reuse rather than correctness.

- **Hoist the controller-build block.** `teacache_controller = None / if enable_teacache: build(stage1_steps, thresh); reset()` now exists three times verbatim (`ti2vid_two_stages.py:533`, `ti2vid_two_stages_hq.py:226`, and this PR). A single `_make_stage1_teacache(enable_teacache, stage1_steps, teacache_thresh) -> TeaCacheController | None` on `TI2VidTwoStagesPipeline` (HQ overriding it to use `_build_hq_teacache_controller`) would let a2v inherit the Euler variant for free and give the 2.5 guard one home. Happy to take that as a follow-up if you'd rather keep this PR minimal, but it's a small change.
- **Reuse the existing test scaffolding.** `tests/test_ltx25_two_stage.py` already has `_make_two_stage` (writes a pack config and builds a pipeline through `__init__`), `_EchoLoop` and `_stub_generate`, and `test_teacache_raises_on_25_pack` / `_hq` are the exact shape the a2v guard test needs. The new `__new__`-based fixture bypasses `__init__` and can only prove the `_denoise_stage1` kwarg hop; the `generate_and_save -> _build_teacache_controller -> _denoise_stage1` path, where the bug actually lived, stays untested. An a2v case built on those helpers (stubbing `guided_denoise_loop` in `a2vid_two_stage` and asserting `calls[0]["teacache"] is controller`) would cover both the wiring and the guard, and you could drop `_LoopSpy`.

PR #141, opened 2026-09-19
fix(pipelines): keep the terminal sigma when stage steps are shortened

coeyai:
## What

`stage1_steps` and `stage2_steps` shorten a schedule with `table[: steps + 1]`. That slice drops the terminal `0.0`, so a shortened run never finishes denoising:

| call | sigmas used today | last sigma |
|---|---|---|
| `stage1_steps=3` | `[1.0, 0.99375, 0.9875, 0.98125]` | 0.98125 (almost pure noise) |
| `stage2_steps=2` | `[0.909375, 0.725, 0.421875]` | 0.421875 |
| `stage2_steps=1` | `[0.909375, 0.725]` | 0.725 |

This PR adds `scheduler.shorten_schedule(table, steps, keep="head" | "tail")`, which always ends at the table's last value, and uses it at all 11 call sites:

- **Stage 1 keeps the head** and jumps to 0.0 on the last step. It starts from pure noise, so it has to start at `1.0`.
- **Stage 2 keeps the tail**, the same convention `ic_lora.py` already uses for the control-aware refine (`DISTILLED_SIGMAS[len - 1 - n:]`).

`None`, 0 or a full step count returns the table unchanged, so **default renders are identical**. `test_stage_step_truncation_applies_to_the_25_tables` pinned the old slice and is updated to pin the new tables. `TestShortenSchedule` covers the helper.

## How we found it, and what the fixed tier looks like

We profiled LTX 2.5 (q8 pack) on an M3 Max with 64 GB for [GroundSlate](https://coey.com/software/groundslate). We wanted a fast "audition" tier: stage two in one step instead of three. The library's `stage2_steps=1` returned a noisy latent, so we passed whole tables instead. At 768x512x121, same prompt and seed, `stage2_steps=1` with the tail (`[0.421875, 0.0]`) measured:

- **stage two:** 34.3 s against 98.4 s; the whole render 118.2 s against 183.1 s (1.55x)
- **quality against the three-step render:** SSIM 0.919, LPIPS 0.162, VMAF 9.1. It's the same seed, composition and motion, but the gaze and fine detail are not settled. It's a preview of the shot, not a softer copy of it.
- **the baseline is exact:** two plain renders of that seed decode to identical frames on an idle machine (SSIM 1.000, zero MSE).

The full study, with the rows, event logs and stills, is at https://github.com/coeyai/ltx-2-mlx-speed-study.

## Tests

`ruff check` and `ruff format --check` are clean. `pytest tests --ignore=tests/regression`: 907 passed. The only failures are the 46 `ltx_trainer_mlx` tests, which fail because I didn't install the trainer package in the test venv.


dgrauet (changes requested):
Thanks — good catch, and a clean fix. Diagnosis confirmed: `table[: steps + 1]` is a local addition of this port (upstream's distilled pipeline takes the sigma tables whole, with no step-count parameter), and it silently dropped the terminal `0.0`, so a shortened run never finished denoising. Verified locally on your branch: 57 scheduler/2.5 tests pass, ruff clean, default renders unchanged on both the 2.3 and 2.5 tables.

**What works well**

- All 11 call sites converted; `None` / `0` / a full step count return the table itself, so nothing changes by default.
- `keep="tail"` for stage 2 matches the convention `ic_lora --refine-steps` already uses (`DISTILLED_SIGMAS[len - 1 - n:]`), and `start_sigma = sigmas_2[0]` keeps the re-noising consistent. Your measurements on the one-step stage 2 back it up.
- Tests pin the exact shortened tables and the terminal-zero invariant.

**One point to settle: `keep="head"` for stage 1**

The distilled table is front-loaded (five sigmas between 1.0 and 0.975), so "head" spends all the steps on tiny moves and then jumps to 0 in one go:

```
stage1_steps=2 -> [1.0, 0.99375, 0.0]
stage1_steps=4 -> [1.0, 0.99375, 0.9875, 0.98125, 0.0]
```

That ends at zero (bug fixed) but it is effectively a one-step generation from σ≈0.98, and it is the one variant the study does not measure. Since stage 1 must start at 1.0 (`create_noised_state(sigma=1.0)`), a shortening that keeps the table's large steps would be the pinned-start tail:

```python
if keep == "head":
    return [table[0], *table[-steps:]]      # e.g. steps=3 -> [1.0, 0.725, 0.421875, 0.0]
```

Neither variant has an upstream reference, so I am fine with either — but "head" as written needs either a measurement showing it produces a usable preview, or the pinned-start tail. Could you pick one and say why in the docstring?

**Minor**

- `keep: str` → `Literal["head", "tail"]`.
- `docs/PIPELINES.md` (the `--distilled`, `--two-stage`, `ic-lora`, `lipdub` rows) and CLAUDE.md should say that `--stage2-steps N` now takes the last N sigmas of the table — it changes what existing `--stage2-steps 2` users get (start at 0.725 instead of 0.909), which is the intended fix but worth a line in the PR body / changelog.
- The branch is behind `main` (0.15.7 was released today); it will need an update before the squash merge.

The trainer test failures you mention are just the missing `ltx-trainer` package in your venv (`uv sync --all-extras` installs it); CI runs them.

